PR #11, opened 2015-01-27
Represent and partition netlist Nets.

mundya:
A (basic) net has a source, a weight and a set of sinks.

Nets can be partitioned by replacing sinks or sources by either single
objects or lists of replacements.  See the docstring for
`rig.netlist.Net.partition`.


mossblaser:
Looks good to me (though I have not verified functioning of code!)

PR #15, opened 2015-02-11
Implement fixpoint conversions.

### LINKED ISSUE #14: Migrate fixed point code
Migrate fixed point conversion code from Nengo/SpiNNaker.  Ensure easy use with `MatrixRegion`.

mundya:
Fixes #14.

 - `float_to_fix` can convert individual values to fixpoint.
 - `FixPointFormatter` can be used with `MatrixRegion` types to convert entire
    matrices of values to fixpoint.
 - `fix_to_float` can convert individual values from fixpoint.

TODO:
 - Provide a NumPy equivalent of `fix_to_float`.

Comments welcomed!

### DIFF
--- a/rig/fixed_point.py
+++ b/rig/fixed_point.py
@@ -1,0 +1,127 @@
+"""Fixed point conversion utilities.
+"""
+import numpy as np
+
+
+def float_to_fix(value, signed, n_bits, n_frac):
+    """Convert a floating point value to a fixed point value.
+
+    Parameters
+    ----------
+    value : float
+        The value to convert.
+    signed : bool
+    n_bits : int
+    n_frac : int
+    """
+    mask = int(2**n_bits - 1)
+    min_v, max_v = validate_fp_params(signed, n_bits, n_frac)
+
+    # Saturate values
+    value = np.clip(value, min_v, max_v)
+
+    if value < 0:
+        fp_val = (1 << n_bits) + int(value * 2**n_frac)
+    else:
+        fp_val = int(value * 2**n_frac)
+
+    assert 0 <= fp_val < 1 << (n_bits + 1)
+    return fp_val & mask
+
+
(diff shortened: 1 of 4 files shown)


mossblaser:
* Presumably `FixPointFormatter` will be passed to regions of various sorts? If so, how does one deal with conversion in the opposite direction?
* Being picky but isn't this better named along the lines of numpy_to_fix(?)
* Otherwise looks good :D (though I have not actually tested that it is correct or not ;) I'll trust your tests.)

mundya:
> - Presumably FixPointFormatter will be passed to regions of various sorts? If so, how does one deal with conversion in the opposite direction?

Is the todo :smile: 

I don't think it's worth renaming, `float_to_fix` works with Python floats and the formatter is really for the MatrixRegions, so I think it's safe to leave that as named..?

mossblaser:
It's not unsafe per-se but the alternative name tells you what it does (i.e. it really is just a type conversion function) the `FixPointFormatter` name implies something somewhat more grand and also disassociated with the likes of `float_to_fix`. Just personal taste though ;).

mossblaser:
Summary of offline discussion:

* All conversion things will come to look like `FixPointFormatter` -- (the TODO should still be considered!)
* Suggested alternative name along lines of `NumpyToFixConverter`
* Jamie called pedantry.

Additional thoughts:

If the module gets renamed more generically "data_converters" the names drop to "NumpyToFix" and "FloatToFix" etc. which is fairly pleasing to my ears.(??)

mundya:
Awesome, thank you :smile_cat:

I'll get on this tomorrow.

mossblaser:
Perhaps a better name for the middle would be type_casts?

mundya:
So I think this does most of what we discussed?

mossblaser:
LGTM!

PR #16, opened 2015-02-12
Refine regions

mundya:
From discussions with @mossblaser 
Simplifies the design of the regions utilities.


mossblaser:
New API etc: looks great to me!

On to the pedantry: Despite comments and some implementation effort implying generality the Keyspace region only really supports keyspaces between 1-16, 24-32 bit keys (inclusive) due to use of struct for packing. SpiNNaker itself, of course, only "really" supports 32-bit keys. Personal preference would be to wap an assert to check keyspaces are <= 32 bits and always use a whole word per entry.

mundya:
Yeah, that sounds sensible.

mundya:
How about just asserting ==32?  That seems no less useful than <= 32.

mossblaser:
Agreed :)

mundya:
Done.

PR #17, opened 2015-02-16
Read SCAMP struct files.

mundya:
Reads struct files.

I'll implement doing fancy things with this in a separate PR if that's alright?


rig/machine_control/struct_file.py
```diff
@@ -0,0 +1,122 @@
...
+    structs = dict()
+
+    # Holders for the current struct
+    name = None
+
+    # Iterate over every line in the file
+    for i, l in enumerate(fp):
+        # Empty the line of comments, if the line is empty then skip to the
+        # next line.  Split on whitespace to get the tokens.
+        tokens = re_comment.sub(b"", l).strip().split()
```
mossblaser: Why all the byte-strings? (/me is not familiar!)

rig/machine_control/struct_file.py
```diff
@@ -0,0 +1,122 @@
...
+        if len(tokens) == 0:
+            continue
+        elif len(tokens) == 3:
+            # 3 tokens implies header data
+            (key, _, value) = tokens
+
+            if key == b"name":
+                if name is not None:
```
mossblaser: This check won't be carried out for the last struct, will it?

rig/machine_control/struct_file.py
```diff
@@ -0,0 +1,122 @@
...
+import re
+
+
+def read_struct_file(fp):
+    """Read a struct file defining the location of variables in memory.
+
+    Parameters
+    ----------
+    fp : file
+        A file-object containing the struct file to read.
```
mossblaser: Could do with:

    Returns
    -------
    {struct_name: :py:class:`~rig.machine_control.Struct`}

rig/machine_control/struct_file.py
```diff
@@ -0,0 +1,122 @@
...
+def num(value):
+    """Convert a value from one of several bases to an int."""
+    if re_hex_num.match(value):
+        return int(value, base=16)
+    else:
+        return int(value)
+
+
+class Struct(object):
+    """Represents an instance of a struct."""
```
mossblaser: Attributes
    ----------
    name : str
        Name of the struct
    size : int
        Total size of the struct in bytes.
    base : int
        Base address of the struct in memory.
    fields : {field_name: :py:class:`~rig.machine_control.StructField`}
        Definitions of fields in the struct.

rig/machine_control/tests/test_struct_file.py
```diff
@@ -0,0 +1,160 @@
...
+def test_num(s, val):
+    assert num(s) == val
+
+
+@pytest.fixture
+def temp_struct_file(data):
+    """Writes the test data into a temporary file to be tested against.
+    """
+    # Create the temporary file and write in the data
+    fp = tempfile.TemporaryFile()
```
mossblaser: For what its worth, you may enjoy "StringIO" which makes a file-like out of a string.

https://docs.python.org/2/library/stringio.html

Also, doesn't this fixture leak temporary files?

mossblaser:
Looks good to me minus the possibly not checking the last struct's size etc. is set.

Otherwise just the usual nit-picky crap which you're free to ignore :)

rig/machine_control/struct_file.py
mundya: This is because of fun Python2/3 changes (default strings in Python 2 = bytes, in Python 3 = unicode)...  I'm open to suggestions!

rig/machine_control/struct_file.py
mundya: Fixed.

rig/machine_control/struct_file.py
mundya: Fixed.

rig/machine_control/struct_file.py
mundya: Fixed.

mundya:
I'll squash before merging if this is ok.

PR #18, opened 2015-02-16
Initial Place and Route Algorithm Implementations

mossblaser:
TODOs:

* [x] ~~Improve performance of Hilbert placer~~ (performance is poor due to dumb breadth-first traversal implementation, more useful to pursue better placement algorithms)
* [x] ~~Documentation clean-up and tutorial/example~~ Postponed 'til general documentation cleanup.
* [x] "Common case" wrapper
* [x] Quick integration test for full process (just the common case wrapper test)
* Utility functions for getting things onto the machine
  * [x] Routing table generation
  * [x] Application loading


rig/tests/par/test_wrapper.py
```diff
@@ -0,0 +1,96 @@
...
+        # Simplest possible case of an empty system
+        m = Machine(1, 1)
+        placements, allocations, application_map, routing_tables = \
+            par({}, {}, [], {}, m)
+        assert placements == {}
+        assert allocations == {}
+        assert application_map == {}
+        assert routing_tables == {}
+
+    def test_ring(self):
```
mossblaser: Highlight for @mundya: This test contains a full run-through of the place/route process for a simple graph.

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+
+    @classmethod
+    def from_vector(cls, vector):
+        """Given a vector from one node to a neighbour, get the link direction.
+
+        Note that any vector whose magnitude in any given dimension is greater
+        than 1 will be assumed to use a machine's wrap-around links.
+
+        Note that this method assumes a system larger than 2x2 (if a smaller
+        system is provided, the function may produce unexpected or imballenced
```
mundya: sp. imbalanced though I'd prefer a more descriptive term.

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+        results).
+
+        Argument
+        --------
+        vector : (x, y)
```
mundya: That's `tuple : (x, y)` surely?

mossblaser: The argument is called `vector` (because it is semantically, though not numpily, a vector). I was hoping `(x, y)` sufficiently implies that the type is a tuple? Is there a a method of specifying this which would be clearer?

mundya: Yeah, I'm silly and forgot numpydoc syntax.

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+        ------
+        :py:class:`~rig.machine.Links`
+            The link direction to travel in the direction indicated by the
+            vector.
+        """
+        x, y = vector
+        if abs(x) > 1:
+            x = -1 if x > 0 else 1
+        if abs(y) > 1:
+            y = -1 if y > 0 else 1
```
mundya: As discussed I think this needs documenting.

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+    (-1, -1): Links.south_west,
+}
+
+
+"""Resource identifier for (monitor and application) processor cores.
+
+Note that this identifer does not trigger any kind of special-case behaviour in
+library functions. Users are free to define their own alternatives.
+"""
+Cores = sentinel.create("Cores")
```
mundya: Should this be singular?

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+SRAM = sentinel.create("SRAM")
+
+
+class Machine(object):
+    """Defines the resources available in a SpiNNaker machine.
+
+    This datastructure makes the assumption that in most systems almost
+    everything is uniform and working.
+
+    This data-structure intends to be completely transparent. Its contents is
```
mundya: Mega nit picks! :smile: "Its contents _are_"

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+        The width of the system in chips: chips will thus have x-coordinates
+        between 0 and width-1 inclusive.
+    height : int
+        The height of the system in chips: chips will thus have y-coordinates
+        between 0 and height-1 inclusive.
+    chip_resources : {resource_key: requirement, ...}
+        The resources available on chips (unless otherwise stated in
+        `chip_resource_exceptions). `resource_key` must be some unique
+        identifying object for a given resource. `requirement` must be a
+        positive numerical value.
```
mundya: Could this have an example `e.g., {Cores: 17}`?

rig/machine.py
```diff
@@ -0,0 +1,220 @@
...
+
+        self.dead_chips = dead_chips.copy()
+        self.dead_links = dead_links.copy()
+
+    def copy(self):
+        """Produce a copy of this datastructure."""
+        return Machine(
+            self.width, self.height,
+            self.chip_resources, self.chip_resource_exceptions,
+            self.dead_chips, self.dead_links)
```
mundya: This only produces a shallow-copy, what's the expected use-case?

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+Placers have the function prototype::
+
+    place(vertices_resources, nets, machine, constraints, **kwargs)
+        -> placement
+
+Where:
+
+* `vertices_resources` is a dictionary `{vertex: resources, ...}` which
+  enumerates the resources required by every vertex to be placed. `vertex`
+  should be a unique object which sensibly implement `__hash__` and `__eq__`.
```
mundya: sensibly implements

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+Where:
+
+* `vertices_resources` as defined above.
+* `nets` as defined above.
+* `machine` as defined above.
+* `constraints` as defined above.
+* `placement` is a dictionary of the format returned by a placer. Note that
+  this placement must be valid (i.e. no vertices on dead/non-existant chips):
+  failiure to comply with this requirement will result in undefined behaviour.
```
mundya: sp. failure

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+Where:
+
+* `vertices_resources` as defined above.
+* `nets` as defined above.
+* `machine` as defined above.
+* `constraints` as defined above.
+* `placements` is a dictionary of the format returned by a placer. Note that
+  this placement must be valid (i.e. no vertices on dead/non-existant chips):
+  failiure to comply with this requirement will result in undefined behaviour.
```
mundya: sp. failure

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+* Most users will wish to use a ReserveResourceConstraint to indicate that core
+  0 is in use as a monitor processor and cannot be allocated.
+
+* One can allocate zero, one or multiple cores at once, an ability which should
+  be treated with some care (see below).
+
+* Routers *do* in fact care about cores and thus, in order to produce a routing
+  with routes which terminates in cores, there must exist exactly one resource
+  which mapps 1:1 to cores on a chip. Note that if this is not done, routers
```
mundya: sp. Unless mapps expands to "maps and only maps"? :worried:

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+  still produce routes which connect all involved chips but these routes simply
+  wont terminate with a core.
```
mundya: won't

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+
+As a result, just as a vertex may consume more than one byte of memory, it may
+also consume zero, one or more cores.  However, since a vertex is assigned to
+exactly one chip, it cannot consume more cores than are available on a single
+chip: vertices are never be split across chips.  If an application requires
```
mundya: "can never be" or "are never"

rig/place_and_route/__init__.py
```diff
@@ -0,0 +1,177 @@
...
+this type of behaviour, users must perform this step in an application-defined
+preprocessing step before placement.
```
mundya: I don't think "preprocessing" is necessary given that you say "before".

rig/place_and_route/constraints.py
```diff
@@ -0,0 +1,125 @@
...
+
+class AlignResourceConstraint(object):
+    """Force alignment of start-indices of resource ranges.
+
+    For example, this can be used to ensure assignments into SDRAM are word
+    aligned.
+
+    Note: placers are not obliged to be aware of or compensate for wastage of a
+    resource due to this constraint and so may produce impossible placements
+    when in the even of large numbers of individual items using a non-aligned
```
mundya: ~~when~~ in the event

rig/place_and_route/constraints.py
```diff
@@ -0,0 +1,125 @@
...
+    __slots__ = ["resource", "alignment"]
+
+    def __init__(self, resource, alignment):
+        self.resource = resource
+        self.alignment = alignment
+
+
+class RouteEndpointConstraint(object):
+    """Force the endpoint of a path through the network to be a particular
+    route.
```
mundya: Is this is really a link constraint? It forces routes to end on a given link, whereas the docstring talks about routes.

rig/place_and_route/route/ner.py
```diff
@@ -0,0 +1,436 @@
...
+from ..constraints import RouteEndpointConstraint
+
+from ...machine import Links, Cores
+
+from ...routing_table import Routes
+
+from ..routing_tree import RoutingTree
+
+
+def concentric_hexagons(radius, start=(0, 0)):
```
mundya: This feels like it's of sufficient utility that it should be exposed elsewhere: a geometry module?

rig/place_and_route/util.py
```diff
@@ -0,0 +1,105 @@
...
+        chip_cores.update(range(core_slice.start, core_slice.stop))
+
+    return application_map
+
+
+def build_routing_tables(routes, net_keys):
+    """Convert a set of RoutingTrees into a per-chip set of routing tables.
+
+    This command produces routing tables with entries ommitted when the route
+    does not change direction.
```
mundya: sp omitted.

It's also useful to generate routing tables with default routes explicitly marked for further minimisation (or if, for any reason, default routing is turned off).  Could this be an argument to this function?

rig/place_and_route/constraints.py
mundya: Ok, this fits the `Routes` enum so I'm happy with this provided that it's clear which type of route is being referred to.  English has too few synonyms.

mundya:
LGTM; I've not read through the tests because I'm reckless.

rig/machine.py
mossblaser: The copying is done in the constructor (/me is copying the convention you may have started in the Net container :)).

The use case is that some constraints (e.g. reserved resource constraints) are easily modelled to a placer by simply reducing the total resources available by the reserved amount. The most straight-forward thing to do would be to copy the Machine and modify it accordingly rather than having to check if a constraint exists and if not, looking at the machine to determine resource availability.

rig/place_and_route/__init__.py
mossblaser: Haha :P, no just a typo.

rig/place_and_route/constraints.py
mossblaser: Good catch, this one is a pain in the arse since it originally was called a Link constraint (and was on links) but it would be legitimately useful to constrain to other endpoints too, e.g. the monitor core.

Some minor re-wording will occur which may/may not improve the situation slightly.

rig/place_and_route/route/ner.py
mossblaser: There are quite a large number of geometry-related utility functions floating around. Other pertinent examples are the shortest-path and distance calculation functions. That and "hey tell me how a board is laid out" for example.

I suppose a root-level geometry module would be appropriate. I'll flag this for refactoring soon but not this PR.

rig/place_and_route/util.py
mossblaser: I presume this is the case for your minimisation algorithm? If so I'd be happy to add the option :).

Re default routing being turned off, it can't be, minus someone adding a globally matching route for drop packets to the end of the routing table...

rig/machine.py
mundya: Cool, sorry.

rig/place_and_route/route/ner.py
mundya: :+1:

rig/place_and_route/util.py
mundya: > I presume this is the case for your minimisation algorithm? If so I'd be happy to add the option :).

It is - I forgot that needed to move here one day, maybe.

> minus someone adding a globally matching route for drop packets to the end of the routing table...

Which, ofcourse, could never happen by accident :worried:

mundya:
Merge when ready sir.

PR #29, opened 2015-02-26
Refactor of SCP code

mossblaser:
- [x] Application loading
  - [x] Generating regions
    - [x] Use lazy tree structure  -- __AM__: I don't remember this being a _requirement_
    - [x] Have insane amounts of fun doing so -- This, of course, is
  - [x] Loading from sets of applications
  - [x] **IMPORTANT** Test APLX and test suite with live board
     - [x] ~~Sort !@~#& Makefile hell.~~
     - [x] Test for signals - using `load_application` kind of demands this.
     - [x] Check that application is loaded correctly - now part of `load_application`, should be good once #39 is merged.
     - [x] ~~Test diagnostics... chicken and egg problem? (JH: surely this is a sort-of integration test?)~~
- [x] Send `sver` to system to determine maximum SDP packet size when first connected
- [x] Signal applications
 - [x] Change state
 - [x] ~~Query state~~ (Now #35 )
- [x] Get process listing (i.e. ybug's `ps`) (#40)
- [x] Load and retrieve router table entries
 - [x] Load
 - [x] ~~Retrieve~~ (Now #37)
- [x] Utility method for _reading_ struct values.
  - [x] Enable reading the `vcpu` struct for a specified core correctly (#40)
- [x] Construct a `Machine` describing the machine's integrity (#36)
  - [x] Implementation
  - [x] Mock-test
  - [x] Live machine test


rig/machine_control/__init__.py
```diff
@@ -1,2 +1,36 @@
+"""A collection of utilities for controlling the operation of a SpiNNaker
+machine.
+
+SpiNNaker software can be controlled through the use of the SpiNNaker Control
+Protocol (SCP) and hardware can be controlled through the use of the BMP
+system.  A :py:class:`~MachineController` wraps up these interfaces and
```
mossblaser: "SpiNNaker software and board management processors (BMPs) can be controlled through the use of the SpiNNaker Control Protocol (SCP)."

rig/machine_control/__init__.py
```diff
@@ -1,2 +1,36 @@
...
+provides higher-level methods to fulfil various machine-management tasks.
+Further classes are provided to manipulate SpiNNaker Datagram Protocol (SDP)
+and SCP packets directly, and a stop-and-wait implementation of SCP.
+
+General usage
+=============
+
+Often it will be sufficient to instantiate and use a
+:py:class:`~MachineController` to boot a SpiNNaker machine and control the
+execution of applications upon it.  The MachineController will convert
```
mossblaser: "The MachineController converts" -- a bit more certainty is due! ;)

rig/machine_control/__init__.py
```diff
@@ -1,2 +1,36 @@
...
+high-level commands into SCP packet(s) which will be multiplexed over as many
```
mossblaser: "SCP packet(s) which *are* be multiplexed"

rig/machine_control/__init__.py
```diff
@@ -1,2 +1,36 @@
...
+ethernet connections as it is aware of.  As the SCP implementation used is
+blocking latency is best reduced by ensuring that SCP packets enter the
```
mossblaser: comma: "blocking**,** latency is best reduced"

rig/machine_control/__init__.py
```diff
@@ -1,2 +1,36 @@
...
+SpiNNaker machine at the ethernet connection nearest their destination chip.
+
+Representing SCP and SDP packets
+================================
+
+SCP and SDP packets can be extracted from strings of :py:class:`bytes` received
+over the network or constructed into bytes for transmission.
```
mossblaser: s/constructed/assembled/ (serialised?)

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
+"""Boot constructs for a SpiNNaker machine.
+
+Implementation is reconstructed from a Perl implementation which forms a
```
mossblaser: .. warning::
    ...

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+from . import struct_file
+import socket
+import struct
+import time
+
+# Specifies the size of packets that should be sent to SpiNNaker to boot the
+# board, and to which port.
+BOOT_MAX_BLOCKS = 32  # 32 kiB limit in DTCM
+BOOT_WORD_SIZE = 256
+BOOT_BYTE_SIZE = 1024
```
mossblaser: These guys aren't quite clear. Isn't one a function of the other, anyway so:

    BOOT_BYTE_SIZE = 1024  # Size of the boot image in bytes
    BOOT_WORD_SIZE = BOOT_BYTE_SIZE / 4

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+PORT = 54321
+BOOT_DELAY = 0.01  # Pause between transmitting boot packets
+
+# Specifies where in the boot data the struct containing default values should
+# be written, and how many bytes of it should be written.
+BOOT_DATA_OFFSET = 3*128
+BOOT_DATA_LENGTH = 128
+
+spin1_boot_options = {
```
mossblaser: # Boot options for different kinds of single-board SpiNNaker systems.

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    hostname : str
+        Hostname or IP address of the SpiNNaker board to boot.
+    width : int
+        Width of the machine (0 < w < 256)
+    height : int
+        Height of the machine (0 < h < 256)
+    cpu_frequency : int
+        CPU clock-frequency.  **Note**: The default (200 MHz) is known
+        safe.
+    hardware_version : int
```
mossblaser: "Version number of the SpiNNaker boards used in the system (e.g. SpiNN-5 boards would be 5). At the time of writing this value is ignored and can be safely set to the default value of 0."

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    led_config : int
```
mossblaser: "Defines LED pin numbers for the SpiNNaker boards used in the system. The four least significant bits (`3:0`) give the number of LEDs. The next four bits give the pin number of the first LED, the next four the pin number of the second LED, and so forth. At the time of writing, all SpiNNaker board versions have their first LED attached to pin 0 and thus the default value of 0x00000001 is safe."

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    boot_data : bytes
+        Data to boot the machine with.
+    struct_data : bytes or None
+        Data to interpret as a representation of the memory layout of data
+        within the boot data.
+
+    Notes
+    -----
+    The constants `rig.boot.spinX_boot_options` can be used to specify boot
```
mossblaser: `rig.machine_control.boot.spinX_boot_options`

That said, I would prefer such constants lived in that `boot` module...

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    parameters, for example:
```
mossblaser: `for example::` (grr ReST...)

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def boot(hostname, width, height, cpu_frequency=200, hardware_version=0,
+         led_config=0x00000001, boot_data=None, struct_data=None):
+    """Boot a SpiNNaker machine of the given size.
+
+    Parameters
+    ----------
+    hostname : str
+        Hostname or IP address of the SpiNNaker board to boot.
```
mossblaser: "Hostname or IP address of the SpiNNaker chip to boot as (0, 0)." (i.e. chip not board!)

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+                             hw_ver=hardware_version,
+                             cpu_clk=cpu_frequency,
+                             led0=led_config,
+                             unix_time=int(time.time()),
+                             boot_sig=int(time.time()),
+                             root_chip=1)
+    struct_packed = sv.pack()
+
+    # NOTE I am assured by ST that writing into boot data bytes 384 onwards
+    # is unlikely to change in the foreseeable future.
```
mossblaser: Move this disclaimer up to the definition of `BOOT_DATA_OFFSET`.

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    buf = bytearray(boot_data)
+    buf[BOOT_DATA_OFFSET:BOOT_DATA_OFFSET+BOOT_DATA_LENGTH] = \
+        struct_packed[:BOOT_DATA_LENGTH]
```
mossblaser: If `len(struct_packed)` is < `BOOT_DATA_LENGTH` then won't this shift everything after this data in the binary? Wouldn't this be Very Bad (TM)?

If `len(struct_packed)` is > `BOOT_DATA_LENGTH` then surely some bits of the struct are going to randomly get chopped off (possibly even mid-way through a value). Isn't this also Very Bad (TM)?

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    assert len(buf) < 32768  # Assert that we fit in DTCM
```
mossblaser: Magic number! This appears to live in the SV struct so perhaps it could be loaded from there? Also would it be clearer to write 32 * 1024?

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+
+    # Create a socket to communicate with the board
+    sock = socket.socket(type=socket.SOCK_DGRAM)
+    sock.connect((hostname, PORT))
+
+    # Transmit the boot data as a series of SDP packets.  First determine
+    # how many blocks must be sent and transmit that, then transmit each
+    # block.
+    n_blocks = len(buf) // BOOT_BYTE_SIZE + (
+        1 if len(buf) % BOOT_BYTE_SIZE != 0 else 0)
```
mossblaser: `(len(buf) + BOOT_BYTE_SIZE - 1) // BOOT_BYTE_SIZE`

Isn't that the more conventional way to get the ceiling truncation?

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+
+
+def boot_packet(sock, cmd, arg1=0, arg2=0, arg3=0, data=b""):
+    """Create and transmit a packet to boot the machine.
+
+    Parameters
+    ----------
+    sock : :py:class:`~socket.socket`
+        Connected socket to use to transmit the packet.
+    cmd : int
```
mossblaser: Can these be named/Enum'd?

    class BootCommand(IntEnum):
        """Boot packet command numbers"""
        
        start = 1
        send_block = 3
        finish = 5

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+        boot_packet(sock, 3, a1, data=data)
+        time.sleep(BOOT_DELAY)
+        block += 1
+
+    # Send the END command
+    boot_packet(sock, 5, 1)
+
+    # Close the socket and give time to boot
+    sock.close()
+    time.sleep(2.0)
```
mossblaser: Magic number. `POST_BOOT_DELAY` maybe?

rig/machine_control/boot.py
```diff
@@ -0,0 +1,160 @@
...
+    arg3 : int
+    data : :py:class:`bytes`
+        Optional data to include in the packet.
+    """
+    PROTOCOL_VERSION = 1
+
+    # Generate the (network-byte order) header
+    header = struct.pack("!H4I", PROTOCOL_VERSION, cmd, arg1, arg2, arg3)
+
+    # Format the data from little- to network-/big-endian
```
mossblaser: Comment should be on the while loop(?)

rig/machine_control/consts.py
```diff
@@ -0,0 +1,48 @@
...
+
+
+class LEDAction(enum.IntEnum):
+    """Indicate the action that should be applied to a given LED."""
+    on = 3
+    off = 2
+    toggle = 1
+
+
+class IPTagCommands(enum.IntEnum):
```
mossblaser: Docstring?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+    Attributes
+    ----------
+    """
+    def __init__(self, initial_host, n_tries=5, timeout=0.5, app_id=None,
+                 boot_data=None, struct_data=None):
+        """Create a new controller for a SpiNNaker machine.
+
+        Parameters
+        ----------
+        initial_host : string
```
mossblaser: "Hostname or IP address of the SpiNNaker chip to connect to. If the board has not yet been booted, this will become chip (0, 0)."

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        n_tries : int
```
mossblaser: "Number of SDP packet retransmission attempts."

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        timeout : float
+        app_id : int or None
```
mossblaser: Presumably this will disappear into the context manager magic?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        boot_data : bytes or None
```
mossblaser: Should this not be an argument to boot?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            Data to boot the machine with.  If None then the default boot file
+            provided with rig will be used.
+        struct_data : bytes or None
```
mossblaser: Ah, so sv isn't read from the machine but rather the version pumped into the boot file is used? Also, isn't this a string (i.e. the contents of sark.struct, an ASCII formatted file)?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # Create the initial connection
+        self.connections = [
+            SCPConnection(initial_host, n_tries, timeout)
+        ]
+
+    def __call__(self, **context_args):
+        """Create a new context for use with `with`.
+
+        E.g:
```
mossblaser: E.g:: (rst...)

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            with controller(x=3, y=4):
+                # All commands will now communicate with chip (3, 4)
+        """
+        return self.get_new_context(**context_args)
+
+    def send_scp(self, x, y, *args, **kwargs):
```
mossblaser: Shouldn't this use contextual arguments for x and y?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            Height of the machine (0 < h < 256)
+
+        For further boot arguments see
+        :py:func:`~rig.machine_controller.boot.boot`.  **Regardless** of the
+        values passed to this method the struct and boot data from `__init__`
+        will be used.
+
+        Notes
+        -----
+        The constants `rig.boot.spinX_boot_options` can be used to specify boot
```
mossblaser: The reference here is out of date: these appear to live in rig.machine_control.boot

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        Will boot the Spin5 board connected to by `controller`.
+        """
+        boot_kwargs.update({
+            "boot_data": self.boot_data,
+            "struct_data": self.struct_data,
+        })
+        boot.boot(self.initial_host, width=width, height=height, **boot_kwargs)
+
+    @ContextMixin.use_contextual_arguments
+    def software_version(self, x=Required, y=Required, processor=0):
```
mossblaser: get_software_version? sver?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+        # arg2 => version number and buffer size
+        version = (sver.arg2 >> 16) / 100.
+        buffer_size = (sver.arg2 & 0xffff)
+
+        return CoreInfo(p2p, pcpu, vcpu, version, buffer_size, sver.arg3,
+                        sver.data)
+
+    @ContextMixin.use_contextual_arguments
+    def write(self, address, data, x=Required, y=Required, p=Required):
```
mossblaser: Is p=0 not a sensible default here? (e.g. when writing to SDRAM, the common case I would guess, there is no reason to do so via an application core?)

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        address : int
+            The address at which to start writing the data.
+        data : :py:class:`bytes`
+            Data to write into memory.
+        """
+        # While there is still data perform a write: get the block to write
+        # this time around, determine the data type, perform the write and
+        # increment the address
+        while len(data) > 0:
+            block, data = data[:256], data[256:]
```
mossblaser: Magic numbers!

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            dtype = address_length_dtype[(address % 4, len(block) % 4)]
+            self._write(x, y, p, address, block, dtype)
+            address += len(block)
+
+    def _write(self, x, y, p, address, data, data_type=DataType.byte):
+        """Write a bytestring to an address in memory.
```
mossblaser: "Write a SCP command's worth of data."

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            Data to write into memory.  Must be <= 256 bytes.
+        data_type : :py:class:`~rig.machine_controller.consts.DataType`
+            The size of the data to write into memory.
+        """
+        length_bytes = len(data)
+        self.send_scp(x, y, p, SCPCommands.write, address, length_bytes,
+                      int(data_type), data, expected_args=0)
+
+    @ContextMixin.use_contextual_arguments
+    def read(self, address, length_bytes, x=Required, y=Required, p=Required):
```
mossblaser: Ditto comments about p being defaulted.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            # Determine the data type to use
+            dtype = address_length_dtype[(address % 4, reads % 4)]
+
+            # Perform the read and increment the address
+            data += self._read(x, y, p, address, reads, dtype)
+            address += reads
+
+        return data
+
+    def _read(self, x, y, p, address, length_bytes, data_type=DataType.byte):
```
mossblaser: "Read an SCP command's worth of data."

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        """Read a bytestring from an address in memory.
```
mossblaser: "Read an SCP command's worth of data."

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+            The data is read back from memory as a bytestring.
+        """
+        read_scp = self.send_scp(x, y, p, SCPCommands.read, address,
+                                 length_bytes, int(data_type),
+                                 expected_args=0)
+        return read_scp.data
+
+    @ContextMixin.use_contextual_arguments
+    def iptag_set(self, iptag, addr, port, x=Required, y=Required):
+        """Set the value of an IPTag.
```
mossblaser: A single sentence in the body of this docstring which approximately hints at what an "IPTag" might be would be very helpful here to those scanning the API.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+        Parameters
+        ----------
+        iptag : int
+            Index of the IPTag to set
```
mossblaser: TODO: addr, port, x, y

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        """Get the value of an IPTag.
+
+        Parameters
+        ----------
+        iptag : int
+            Index of the IPTag to get
+
+        Returns
+        -------
+        :py:class:`IPTag`
```
mossblaser: How come so much more comes back compared with setting an IP tag? (/me is curious)

rig/machine_control/boot.py
mundya: > If len(struct_packed) is < BOOT_DATA_LENGTH then won't this shift everything after this data in the binary? Wouldn't this be Very Bad (TM)?

It shouldn't, because it's indexing into a buffer.  I'm not sure what WILL happen though.

> If len(struct_packed) is > BOOT_DATA_LENGTH then surely some bits of the struct are going to randomly get chopped off (possibly even mid-way through a value). Isn't this Double Plus Very Bad (TM)?

Yes. And yes.

I'll stick an `assert len(boot_data) == BOOT_DATA_LENGTH` in.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        Parameters
+        ----------
+        iptag : int
+            Index of the IPTag to clear.
+        """
+        self.send_scp(x, y, 0, SCPCommands.iptag,
+                      int(consts.IPTagCommands.clear) << 16 | iptag)
+
+    @ContextMixin.use_contextual_arguments
+    def led_on(self, led, x=Required, y=Required):
```
mossblaser: Can these three LED functions not be made into one which takes on/off as an argument (which if not specified toggles)?

rig/machine_control/boot.py
mossblaser: As per discussion:
    >>> buf = bytearray(range(10))
    >>> buf
    bytearray(b'\x00\x01\x02\x03\x04\x05\x06\x07\x08\t')
    >>> buf[0:4] = bytearray(range(2))
    >>> buf
    bytearray(b'\x00\x01\x04\x05\x06\x07\x08\t')
Further discussion says that the struct_packed is always BOOT_DATA_LENGTH long.
As a result, the slicing of struct_packed is unnecessary and misleading(?)
Further an assert on its length would be both informative and a good sanity
check. (Shudders slightly at the documentation situation with this stuff...)

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        ----------
+        led : int
+            Number of the LED to set the state of (0-3)
+        """
+        arg1 = int(consts.LEDAction.toggle) << (led * 2)
+        self.send_scp(x, y, 0, SCPCommands.led, arg1=arg1, expected_args=0)
+
+    @ContextMixin.use_contextual_arguments
+    def sdram_alloc(self, size, tag=0, x=Required, y=Required,
+                    app_id=Required):
```
mossblaser: Ah... now I see why "Required" is necessary. Hmm... difficult.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+    @ContextMixin.use_contextual_arguments
+    def software_version(self, x=Required, y=Required, processor=0):
+        """Get the software version for a given SpiNNaker core.
+        """
+        sver = self.send_scp(x, y, processor, SCPCommands.sver)
+
+        # Format the result
+        # arg1 => p2p address, physical cpu, virtual cpu
+        p2p = sver.arg1 >> 16
```
mossblaser: Unpacking this into a tuple (x, y) would be very welcome here!

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        # Send the packet and retrieve the address
+        rv = self.send_scp(x, y, 0, SCPCommands.alloc_free, arg1, size, tag)
+        if rv.arg1 == 0:
+            # Allocation failed
+            raise SpiNNakerMemoryError(size, x, y)
+        return rv.arg1
+
+CoreInfo = collections.namedtuple(
+    'CoreInfo', "p2p_address physical_cpu virt_cpu version "
+                "buffer_size build_date version_string")
```
mossblaser: A docstring defining approximately what these values are would be appropriate here. (Note: to get namedtuple docs to appear right in sphinx, "Parameters" should be used to document them... Not ideal but still). (Also, I'm assuming you do unpack p2p_address into an (x, y) tuple).

    class CoreInfo(collections.namedtuple(
                   'CoreInfo',
                   "p2p_address physical_cpu virt_cpu version "
                   "buffer_size build_date version_string")):
        """Information returned about a core by sver.
        
        Paramters
        ---------
        p2p_address : (x, y)
            Logical location of the chip in the system.
        physical_cpu : int
            The physical ID of the core. (Not useful to most users).
        virt_cpu : int
            The virtual ID of the core. This is the number used by all high-level
            software APIs.
        version : float
            Software version number. (Major version is integral part, minor version
            is fractional part).
        buffer_size : int
            Maximum supported size (in bytes) of the data portion of an SCP packet.
        build_date : int
            The time at which the software was compiled as a unix timestamp. May be
            zero if not set.
        version_string : string
            Human readable, textual version information split in to two fields by a
            "/". In the first field is the kernal (e.g. SC&MP or SARK) and the
            second the hardware platform (e.g. SpiNNaker).
        """

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+
+class IPTag(collections.namedtuple("IPTag",
+                                   "addr max port timeout flags count rx_port "
+                                   "spin_addr spin_port")):
+    """An IPTag as read from a SpiNNaker machine."""
+    @classmethod
+    def from_bytestring(cls, bytestring):
+        (ip, max, port, timeout, flags, count, rx_port, spin_addr,
+         spin_port) = struct.unpack("4s 6s 3H I 2H B", bytestring[:25])
```
mossblaser: Struct format should start with "<" to force little-endian, standard-sizes and no alignment.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+        # Convert the IP address into a string, otherwise save
+        ip_addr = '.'.join(str(x) for x in struct.unpack("4B", ip))
```
mossblaser: "<" prefix

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,382 @@
...
+
+        return cls(ip_addr, max, port, timeout, flags, count, rx_port,
+                   spin_addr, spin_port)
+
+
+# Dictionary of (address % 4, n_bytes % 4) to data type
+address_length_dtype = {
+    (i, j): (DataType.word if (i == j == 0) else
+             (DataType.short if (i % 2 == j % 2 == 0) else
+              DataType.byte)) for i in range(4) for j in range(4)
```
mossblaser: Breaking the "for" and aligning with the `(i, j):` would make this far more obvious (to me, anyway) -- it took me a good while to spot!

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
+"""A stop-and-wait blocking implementation of the SCP protocol.
```
mossblaser: ~~stop-and-wait~~

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+"""
+import socket
+from . import packets
+
+
+class SCPConnection(object):
+    """Implements the SCP protocol for communicating with a SpiNNaker machine.
```
mossblaser: "...communicating with a SpiNNaker ~~machine~~ chip."

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+    """
+    error_codes = {}
+
+    def __init__(self, spinnaker_host, n_tries=5, timeout=0.5):
+        """Create a new communicator to handle control of the given SpiNNaker
+        host.
```
mossblaser: "Create a new communicator to handle control of the SpiNNaker chip with the supplied hostname."

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+
+        Parameters
+        ----------
+        spinnaker_host : str
+            A IP address or hostname of the SpiNNaker machine to control.
```
mossblaser: ~~machine~~ chip

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+        n_tries : int
+            The maximum number of tries to communicate with the machine before
```
mossblaser: ~~machine~~ chip

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+            failing.
+        timeout : float
+            The timeout to use on the socket.
+        """
+        # Create a socket to communicate with the SpiNNaker machine
+        self.sock = socket.socket(socket.AF_INET, socket.SOCK_DGRAM)
+        self.sock.settimeout(timeout)
+        self.sock.connect((spinnaker_host, 17893))
```
mossblaser: Magic number

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+        arg2 : int
+        arg3 : int
+        data : bytestring
+        expected_args : int
+            The number of arguments (0-3) that are expected in the returned
+            packet.
+
+        Returns
+        -------
+        :py:class:`~rig.communicator.SCPPacket`
```
mossblaser: Path is now wrong.

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+        # Repeat until a reply is received or we run out of tries.
+        n_tries = 0
+        while n_tries < self.n_tries:
+            # Transit the packet
+            self.sock.send(b'\x00\x00' + packet.bytestring)
+            n_tries += 1
+
+            try:
+                # Try to receive the returned acknowledgement
+                ack = self.sock.recv(512)
```
mossblaser: Magic number. (Where defined this should be documented as the smallest power of two large enough to handle the largest packet the SpiNNaker SDP implementation can produce (256+8 bytes).)

rig/machine_control/sdram_file.py
```diff
@@ -0,0 +1,84 @@
+class SDRAMFile(object):
```
mossblaser: File missing introductory docstring.

    """File-like interface wrapper for SDRAM of SpiNNaker chips."""

rig/machine_control/sdram_file.py
```diff
@@ -0,0 +1,84 @@
...
+    def __init__(self, machine_controller, x, y,
+                 start_address=0x60000000,
+                 end_address=0x6fffffff):
```
mossblaser: Magic default numbers! In my (somewhat strongly held) opinion, the defaults should be the 4mb part near the start of SDRAM which is defined as a "free-for-all" (`sv->sdram_base` to `sv->sdram_heap`.). This encourages "modern" behaviour, i.e. using the memory allocator when using blocks of SDRAM.

Bonus points for alternative constructors which look like "malloc" or which take a tag number as argument.

rig/machine_control/sdram_file.py
```diff
@@ -0,0 +1,84 @@
...
+        ----------
+        n_bytes : int
+            A number of bytes to read.
+
+        Returns
+        -------
+        :py:class:`bytes`
+            Data read from SpiNNaker as a bytestring.
+        """
+        assert self.address + n_bytes < self._end_address
```
mossblaser: The standard behaviour of a python file is to produce no output when an EOF is reached and not to throw an exception.

rig/machine_control/sdram_file.py
```diff
@@ -0,0 +1,84 @@
...
+
+    def write(self, bytes):
+        """Write data to the SDRAM.
+
+        Parameters
+        ----------
+        bytes : :py:class:`bytes`
+            Data to write to the SDRAM as a bytestring.
+        """
+        assert self.address + len(bytes) < self._end_address
```
mossblaser: The [standard behaviour of a python file-like](https://docs.python.org/2/library/io.html#io.RawIOBase.write) is to return the number of bytes actually written. I.e. on writing past the end of the range, less bytes should be reported as written than were returned.

rig/utils/contexts.py
```diff
@@ -0,0 +1,117 @@
+"""Utilities for building contextual functions.
+
+The MachineController has a lot of functions which take the same arguments,
```
mossblaser: "As an example..."

rig/utils/contexts.py
```diff
@@ -0,0 +1,117 @@
...
+will lead to cleaner and clearer code.  For example:
+
+    with controller(app_id=32):
+        with controller(x=1, y=1):
+            controller.do_something(...)
+
+        with controller(x=2, y=2):
+            controller.do_something(...)
+
+Is arguably clearer and less prone to silly mistakes than:
```
mossblaser: "In many cases this is arguably clearer and less prone to silly mistakes than:"

rig/utils/contexts.py
```diff
@@ -0,0 +1,117 @@
...
+
+    controller.do_something(x=1, y=1, app_id=32)
+    controller.do_something(x=2, y=2, app_id=32)
+
```
mossblaser: "Though this form is still useful in other circumstances and thus should still be allowed."

rig/utils/contexts.py
```diff
@@ -0,0 +1,117 @@
...
+"""
+import collections
+import inspect
+import sentinel
+from six import iteritems
+
+
+# Allow specifying keyword arguments as required, i.e., they must be satisfied
+# by either the context OR by the caller.
+Required = sentinel.create('Required')
```
mossblaser: Tip: document things like this in a """docstring""" *underneath* the constant and Sphinx will pick it up! Annoyingly you have to use `:py:data:` rather than `:py:const:` to reference it but it does work!

Also, a little motivation would be in order:

    """Allow specifying keyword arguments as required, i.e., they must be satisfied
    by either the context OR by the caller.
    
    This is useful when a method has optional parameters and contextual arguments::
    
        def sdram_alloc(self, size, tag=0, x=Required, y=Required):
            # ...
    """

setup.py
```diff
@@ -12,6 +12,6 @@
     license="GPLv2",
 
     # Requirements
-    install_requires=["numpy>1.6", "six", "enum34"],
+    install_requires=["numpy>1.6", "six", "enum34", "sentinel"],
```
mossblaser: Oops! That has been required by place and route too! By the way why do we have the requirements*.txt files when this exists?

mossblaser:
As usual, loads of super-high anal comments but generally I very much like it! As is convention, I'm playing with fire by not reviewing the tests!

With comments resolved as you see fit (and any outstanding features added) I'd say that is looking good to go! By the way I'm happy to help out implementing any of the comments here as requested ;).

setup.py
mundya: To please Travis.

rig/machine_control/boot.py
mundya: Agreed, this entire file can move then?

mossblaser: Even better idea!

rig/machine_control/machine_controller.py
mundya: In principle these can vary and should be determined by call `get_software_version`... yay!

mossblaser: Can they be determined from sv?

rig/machine_control/machine_controller.py
mundya: The SC&MP code for setting an IPtag just returns the index of the tag...

rig/machine_control/machine_controller.py
mundya: I think SpiNNaker re-orders :worried: I'll check.

rig/machine_control/sdram_file.py
mundya: > Should probably inherit from RawIOBase

I misread this as `RawlOBase`... not sure why.  This sounds sensible.

rig/machine_control/sdram_file.py
mundya: Yes, the defaults should be that.

rig/utils/contexts.py
mundya: > Tip: document things like this in a """docstring""" underneath the constant and Sphinx will pick it up! Annoyingly you have to use :py:data: rather than :py:const: to reference it but it does work!

I did not known this :smile:

rig/machine_control/boot.py
mundya: Done.

rig/machine_control/boot.py
mundya: Check.

rig/machine_control/consts.py
mundya: Check.

rig/machine_control/machine_controller.py
mundya: Done.

rig/machine_control/machine_controller.py
mundya: I don't think so, mostly because this is usually called by methods in the `MachineController`... but I'm open to persuasion.

rig/machine_control/machine_controller.py
mundya: No, that would be too easy!

rig/machine_control/machine_controller.py
mundya: Done (aside for x and y which I think can be documented once in the class docstring?)

rig/utils/contexts.py
mundya: Done.

rig/machine_control/machine_controller.py
mossblaser: I can't think of a reason not? I understand that this is a behind-the-scenes type of method but unless it was actually private it doesn't seem obvious that it shouldn't... After all, if I'm sending SCP packets which the MachineController doesn't support I'll be doing this.

rig/machine_control/machine_controller.py
mossblaser: Hmm... Probably not worth the fuss in practice... I'd personally stick them in a constant with the other protocol-defined constants with a scary comment above it.

rig/machine_control/machine_controller.py
mossblaser: Ah yes, of course!

rig/machine_control/machine_controller.py
mundya: This is not trivial, because the context argument wrapping mucks around with `*args`.  I can continue faffing or I can cheat a little: do you have a preference?

mossblaser: Ahh... I see :( I'd go with the path of least resistance :). Didn't realise it was a pain...

mundya: I tried :smile_cat:

rig/machine_control/regions.py
```diff
@@ -0,0 +1,200 @@
...
+
+    # For each of these cores build the minimal set of regions
+    for core_mask, coordinates in iteritems(cores_to_targets):
+        regions = reduce_regions_hierarchically(coordinates)
+        for r in regions:
+            yield (r, core_mask)
+
+
+class RegionTree(object):
+    """Tree structure for use in minimising regions."""
```
mossblaser: """A tree structure for use in minimising sets of regions.
    
    A tree is defined which reflects the definition of SpiNNaker regions like so:
    The tree's root node represents a 256x256 grid of SpiNNaker chips. This grid is
    broken up into 64x64 grids which are represented by the (level 1) child nodes of
    the root.  Each of these level 1 nodes' 64x64 grids are broken up into 16x16
    grids which are represented by their (level 2) children. These level 2 nodes
    have their 16x16 grids broken up into 4x4 grids represented by their (level 3)
    children. Level 3 children explicitly list which of their sixteen cores are part
    of the region.
    
    If any of a level 2 node's level 3 children have all of their cores selected,
    these level 3 nodes can be removed and replaced by a level 2 region with the
    corresponding 4x4 grid selected. If multiple children can be replaced with level
    2 regions, these can be combined into a single level 2 region with the
    corresponding 4x4 grids selected, resulting in a reduction in the number of
    regions required. The same process can be repeated at each level of the
    hierarchy eventually producing a minimal set of regions.
    
    This data structure is specified by supplying a sequence of (x, y) coordinates
    of chips to be represented by a series of regions using
    :py:meth:`.add_coordinate`. This method minimises the tree during insertion
    meaning a minimal set of regions can be extracted by :py:meth:`.get_regions`
    which simply traverses the tree.
    """

rig/machine_control/regions.py
```diff
@@ -0,0 +1,200 @@
...
+    mask = 0xffff ^ ((4 << shift) - 1)  # in {0xfffc, 0xfff0, 0xffc0, 0xff00}
+    nx = x & mask  # The mask guarantees that bits 1:0 will be cleared
+    ny = y & mask  # The mask guarantees that bits 1:0 will be cleared
+
+    #        sig bits x | sig bits y |  2-bit level  | region select bits
+    region = (nx << 24) | (ny << 16) | (level << 16) | (1 << bit)
+    return region
+
+
+def reduce_regions_hierarchically(chips):
```
mossblaser: This could safely be called `coalesce_regions` or `minimise_regions` since the implementation is immaterial and the effect is a perfect minimisation of the set of regions required to address exactly the given collection of chips.

rig/machine_control/regions.py
```diff
@@ -0,0 +1,200 @@
...
+    """Create a reduced set of regions by minimising a hierarchy tree.
+
+    Parameters
+    ----------
+    chips : {(x, y) ...}
```
mossblaser: Not a set, just an iterable. (Yey generators!)

rig/machine_control/regions.py
```diff
@@ -0,0 +1,200 @@
...
+        A generator which yields 32-bit region codes which minimally cover the
+        set of given chips.
+    """
+    t = RegionTree()
+    for (x, y) in chips:
+        t.add_coordinate(x, y)
+    return t.get_regions()
+
+
+def get_minimal_flood_fills(targets):
```
mossblaser: **Super Pedant Says**: This doesn't produce a *minimal* set of flood fills, just an approximately minimal one. Perhaps `compress_flood_fill_regions`?

rig/machine_control/regions.py
```diff
@@ -0,0 +1,200 @@
+"""SpiNNaker machine regions.
+
+.. todo::
+    Neaten this documentation!
+
+Regions (not to be confused with rig Regions) are used to specify areas of a
+SpiNNaker machine for the purposes of transmitting nearest neighbour packets or
+for determining which chips should be included in any flood-fill of data or
+application loading.
+
```
mossblaser: The region specification system is defined in "Managing Big SpiNNaker Machines" By Steve Temple.

I've asked ST if there is URL for this doc and am awaiting a reply.

rig/machine_control/regions.py
mundya: Can you write all of my documentation? :wink:

rig/machine_control/regions.py
mossblaser: Update from ST: "No, it was really a discussion document and just circulated by email. The current implementation has made it slightly inaccurate." I have suggested making the region part a standalone appnote and ST is going to look into doing this.

For the moment the above ref is going to be the best option.

mundya: The `RegionTree` docstring documents it pretty well now :smile: thank you.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+        If a :py:class:`str` APLX filename is the first argument then the
+        second is assumed to be a dictionary mapping {(x, y): set([cores]),
+        ...}.  Otherwise the return value of
+        :py:func:`~rig.place_and_route.util.build_application_map` may be used
+        directly.
+
+        An `app_id` can be entered as a keyword argument OR from a context.::
+
+            # Either
+            controller.load_aplx(targets, app_id=30)
```
mossblaser: `app_flags` missing from example.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+            address += len(data)
+
+    def _send_ffe(self, pid, app_id, app_flags, cores, fr):
+        """Send a flood-fill end packet."""
+        arg1 = (NNCommands.flood_fill_end << 24) | pid
+        arg2 = (app_id << 24) | (app_flags << 18) | (cores & 0x3fff)
+        self._send_scp(0, 0, 0, SCPCommands.nearest_neighbour_packet,
+                       arg1, arg2, fr)
+
+    @ContextMixin.require_named_contextual_arguments("app_id", "app_flags")
```
mossblaser: I would be tempted to make app_flags have a default of 0. Unless you're being clever, you don't need any flags. (For example, most users probably won't use the wait flag and then send SIG_START because their application already has a soft barrier on SYNC0).

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+    def load_aplx(self, *args, **kwargs):
+        """Load an APLX to a set of application cores.
+
+        If a :py:class:`str` APLX filename is the first argument then the
+        second is assumed to be a dictionary mapping {(x, y): set([cores]),
+        ...}.  Otherwise the return value of
+        :py:func:`~rig.place_and_route.util.build_application_map` may be used
+        directly.
```
mossblaser: This method can be called in either of the following ways::
    
        load_aplx("/path/to/binary.aplx", {(x, y): set([core, ...]), ...})
        load_aplx({"/path/to/binary.aplx": {(x, y): set([core, ...]), ...}})
    
    Note that the latter format is the same format produced by
    :py:func:`~rig.place_and_route.util.build_application_map`.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+
+        An `app_id` can be entered as a keyword argument OR from a context.::
```
mossblaser: Is this general scheme not suitably explained elsewhere?

rig/machine_control/machine_controller.py
mossblaser: .. warning::
        The loading process is extremely likely, but not guaranteed, to succeed.
        This is because the flood-fill packets used during loading are not
        guaranteed to arrive. The effect of this is one of the following:
        
        * Some regions may be included/excluded incorrectly.
        * Some chips will not receive the complete application binary and will
          silently not execute the binary.
        
        As a result, the user is responsible for checking that each core was
        successfully loaded with the correct binary. At present, the two
        recommended approaches to this are:
        
        * The user should check that the correct number of application binaries
          reach their initial barrier (SYNC0), when this facility is used. This
          is not fool-proof but will flag up all but situations where exactly
          the right number, but the wrong selection of cores were loaded. (At
          the time of writing, this situation is not possible but will become
          a concern in future versions of SC&MP.
        * The user can check the process list of each chip to ensure the
          application was loaded into the correct set of cores.

This warning obviously will need to be re-written once checking becomes part of this method.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+    def _send_ffe(self, pid, app_id, app_flags, cores, fr):
+        """Send a flood-fill end packet."""
+        arg1 = (NNCommands.flood_fill_end << 24) | pid
+        arg2 = (app_id << 24) | (app_flags << 18) | (cores & 0x3fff)
+        self._send_scp(0, 0, 0, SCPCommands.nearest_neighbour_packet,
+                       arg1, arg2, fr)
+
+    @ContextMixin.require_named_contextual_arguments("app_id", "app_flags")
+    def load_aplx(self, *args, **kwargs):
+        """Load an APLX to a set of application cores.
```
mossblaser: I'd be tempted to make the "safe" loading of aphlexen a wrapper function which calls this with wait enabled and then counts the waits.

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,770 @@
...
+            # Build the route as a 32-bit value
+            route = 0x00000000
+            for r in entry.route:
+                route |= r
+
+            data += struct.pack("<2H 3I", i, 0, r, entry.key, entry.mask)
+        self.write(buf, data, x, y)
+
+        # Perform the load of the data into the router
+        self._send_scp(
```
mossblaser: **Bug** (?): If more than 16 routing entries are to be loaded the SCP packet payload will grow beyond 256 bytes and thus won't be handled correctly by SC&MP.

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    """States that an application may be in."""
+    # Error states - further information may be available
+    dead = 0
+    power_down = 1
+    runtime_exception = 2
+    watchdog = 3
+
+    # General states
+    init = 4  # Transitory "(hopefully)"
+    ready = 5
```
mossblaser: This state is called "wait" in sark.h. Further I would probably add:

    # Awaiting SIG_START.

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    c_main = 6
```
mossblaser: # Entered c_main

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    run = 7
```
mossblaser: # Running application event loop

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    pause = 10
+    exit = 11
```
mossblaser: # Application returned from c_main

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Awaiting synchronisation (at a barrier)
+    sync0 = 8
+    sync1 = 9
+
+
+class AppSignal(enum.IntEnum):
+    """Signals that may be transmitted to applications."""
+    # General purpose signals
+    init = 0
```
mossblaser: # Load default application (i.e. SARK)

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    power_down = 1
```
mossblaser: # Power down cores.

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    stop = 2
```
mossblaser: # Stop application and cleanup (e.g. free allocated resources)

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    start = 3
```
mossblaser: # Start applications in AppState.wait

(or "AppState.ready" in the current version)

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    power_down = 1
+    runtime_exception = 2
+    watchdog = 3
+
+    # General states
+    init = 4  # Transitory "(hopefully)"
+    ready = 5
+    c_main = 6
+    run = 7
+    pause = 10
```
mossblaser: # Paused by signal AppSignal.pause

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+
+class AppSignal(enum.IntEnum):
+    """Signals that may be transmitted to applications."""
+    # General purpose signals
+    init = 0
+    power_down = 1
+    stop = 2
+    start = 3
+    pause = 6
+    cont = 7
```
mossblaser: # Continue execution after pausing

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+
+
+class AppSignal(enum.IntEnum):
+    """Signals that may be transmitted to applications."""
+    # General purpose signals
+    init = 0
+    power_down = 1
+    stop = 2
+    start = 3
+    pause = 6
```
mossblaser: # Pause execution of an application

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    cont = 7
+    exit = 8
```
mossblaser: # Request that an application terminates itself (leaving it in AppState.exit)

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    timer = 9
```
mossblaser: # Manually trigger a timer interrupt

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+
+    # Barrier synchronisation
+    sync0 = 4
```
mossblaser: # Continue from AppState.sync0

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    sync1 = 5
```
mossblaser: # Continue from AppState.sync1

rig/machine_control/consts.py
```diff
@@ -0,0 +1,179 @@
...
+    AppSignal.usr0: MessageType.multicast,
+    AppSignal.usr1: MessageType.multicast,
+    AppSignal.usr2: MessageType.multicast,
+    AppSignal.usr3: MessageType.multicast,
+
+    AppDiagnosticSignal.AND: MessageType.peer_to_peer,
+    AppDiagnosticSignal.OR: MessageType.peer_to_peer,
+    AppDiagnosticSignal.count: MessageType.peer_to_peer,
+}
+"""Mapping from an :py:class:`AppSignal` to the :py:class:`MessageType`
```
mossblaser: Mapping from an :py:class:`.AppSignal` to the default :py:class:`.MessageType`

(Also, note the added "." in the crossrefs)

mossblaser:
@mundya To save an awful lot of comments and your time, would it be appropriate for me to push things like documentation directly into this branch?

mundya:
@mossblaser Yes please :)

mundya:
Yak-shaving alert: getting a test APLX looks like it will require either (1) providing a Makefile stub in rig, (2) awaiting the moving of some Makefile code from SpyNNaker to spinnaker_tools.  This is because the "make an APLX" part of the Makefile isn't in spinnaker_tools.

(Roughly lines 34-72 of https://github.com/SpiNNakerManchester/sPyNNaker/blob/master/neural_modelling/src/Makefile.depend which I'm reluctant to vendorise because they are sufficiently general I *believe* they should be in spinnaker_tools.)

rig/machine_control/consts.py
mundya: Yeah... there's more depth in this system than I wanted to plumb.  I believe the last 3 are the only ones that are fixed (?) and these are sufficiently different that they could be removed from this dict and placed elsewhere.

mundya:
Getting the Makefiles working wasn't as hard as anticipated but only works if one is on branch `makefile_fix` of SpiNNakerManchester/spinnaker_tools.  Apparently this will be merged soon.

mossblaser:
Hmphh. Can't assign multiple people...

mossblaser:
Apologies! (Hopefully now fixed)

mundya:
Sadly the 5 s post-boot delay makes for _slow_ tests against the machine.

mossblaser:
Perhaps modify the test program to default to overriding it as 2s on the assumption the test system is small? (5s default is a sensible real-world number for larger machines, I am told).

mundya:
Ug, reading the `vcpu` struct doesn't work at all at the moment...

rig/machine_control/scp_connection.py
```diff
@@ -0,0 +1,152 @@
...
+        # Repeat until a reply is received or we run out of tries.
+        n_tries = 0
+        while n_tries < self.n_tries:
+            # Transit the packet
+            self.sock.send(b'\x00\x00' + packet.bytestring)
+            n_tries += 1
+
+            try:
+                # Try to receive the returned acknowledgement
+                ack = self.sock.recv(consts.SCP_RECEIVE_LENGTH)
```
mossblaser: **Bug:** This value is dependent on the maximum packet size (as discovered by the MachineController's sver). This is probably a bit messy but it should almost certainly come from an argument.

mossblaser:
Ug, `s/get_working_machine/get_machine/g` in commit message of  91c7897...

rig/machine_control/scp_connection.py
mundya: Yay circular dependency... one option is to just set it unrealistically big?

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,1054 @@
...
+        unloaded = application_map
+
+        # Try to load the applications, then determine which are unloaded
+        tries = 0
+        while unloaded != {} and tries <= n_tries:
+            tries += 1
+
+            # Load all unloaded applications, then pause
+            self.flood_fill_aplx(unloaded, app_id=app_id, wait=True)
+            time.sleep(0.1)
```
mundya: This constant should, possibly, be a parameter (well documented!)

rig/machine_control/scp_connection.py
mossblaser: According to ST/the spec, sver must always fit in a 256 packet making it a
safe value for bootstrapping.

mundya:
@mossblaser - I await your rubber stamp :sweat_smile:

rig/machine_control/machine_controller.py
```diff
@@ -0,0 +1,1193 @@
...
+        field_name : string
+            Name of the field to read, e.g., `"eth_addr"`
+
+        .. note::
+            The value returned is unpacked given the struct specification.
+
+        .. warning::
+            This feature is only available if this machine controller was used
+            to boot a board OR an appropriate struct definition has been
+            provided.  To do this use, e.g::
```
mossblaser: Do you think it reasonable that perhaps the MachineController should accept a set of structs as an argument which defaults to simply loading the same struct file the boot command defaults to loading? 99.948% of people will never want any different and hopefully the number of people whose scripts will start with "boot the board" should gradually tend towards 0 over time. (/me is happy to add this quickly).

mundya: That sounds reasonable, merge when ready!

mossblaser:
Not sure why coveralls has been flaky lately but at this point, assuming you're happy with my previous two changes then have my stamp of approval:

Rigged

mundya:
Do you not think "rigged" has negative connotations?  I don't think those `pragma`s are necessary but otherwise I'm happy.

mossblaser:
"Rig -- A set of tools for rigging SpiNNaker programs." ;P Also, I find it funny word to use ;).

Once I've got your go/no-go on the `.coveragerc` change I'll push the big button.

mundya:
> "Rig -- A set of tools for rigging SpiNNaker programs." ;P Also, I find it funny word to use ;).

Next tool: Framed - Like `git blame` only more... persuadable

PR #38, opened 2015-03-08
Fixes reading VCPU struct values.

mundya:
This is one option for fixing reading the vcpu struct.  It maintains the
same API that we had before (`read_struct_field`) but as a result hides
some magic... I'm not sure if we're happy with doing it this way.

 - [ ] Run test against real board.


mossblaser:
I think this is a little too much magic for my taste. A thin wrapper
'read_vcpu' with the required logic would be more to my taste (likewise,
users calling read struct I think should be smart enough to not read sv
from anything but p0 (besides, the cost is only a little overhead, the
alternative is magic behaviour).

PR #39, opened 2015-03-09
Fix the VCPU struct problems.

mundya:
- Adds a new `read_vcpu_struct_field` method
 - Modifies `read_struct_field` to accept the default type of string as an
   argument and to convert to bytes as appropriate.

Allows the live tests to boot a board and run applications having checked that
the cores are in a usable state.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -1,4 +1,5 @@
 import collections
+import six
 from six import iteritems
 import socket
 import struct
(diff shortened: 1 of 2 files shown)


mossblaser:
@mundya Aside from the comment update and the `p=Required` I'd say that is good to go :). (The automatic conversion for membership tests is not a big deal ;))

mundya:
See #40, this fixes the above comments.

PR #40, opened 2015-03-09
Processor status

mundya:
Includes #39.

 - Implements VCPU value reading.
 - Implements `ps` command from ybug (`get_processor_status`)


mundya:
So not a complete `ps`, but enough that writing:

```python
for (x, y) in chips:
    for core in cores:
        cn.get_processor_status(x=x, y=y, p=p)
```

would be useful.

mossblaser:
Holy debug info, batman! LGTM :)

PR #49, opened 2015-03-10
Add `loading_routing_tables`

### LINKED ISSUE #47: Load routing table entries like loading applications
We have `load_applications` which accepts the `application_map` created by place and route but we don't have an analogue for routing tables.

Propose adding `load_routing_tables` which is just a thin wrapper around `load_routing_table_entries` but accepts the output of the place and route tools as input.

mundya:
To load multiple routing tables across a SpiNNaker machine.  The routing
table analogue of `load_application`.

Fixes #47


mossblaser:
LGTM!

PR #58, opened 2015-03-12
Correct `MemoryIO.seek` implementation

### LINKED ISSUE #57: `MemoryIO` Absolute Seek not implemented

mundya:
- Adds `from_what` argument to `MemoryIO.seek` as per the Python standard:
   - `0` seeks from the start of the memory region
   - `1` seeks from the current position
   - `2` seeks from the end of the memory region.

It is possible to seek beyond the region, but in this case reads and writes
will abort safely.

Fixes #57.


mossblaser:
Rigged

PR #59, opened 2015-03-12
Make Travis test against real SpiNNaker hardware

mossblaser:
* [x] Non-booting tests working
* [x] Full power-cycle and boot tests working
* [x] Find permanent home for proxy
* [x] Make booting reliable from travis
* [x] Get proxy permanently setup on CSPC276
* [x] Await a successful few travis runs
* [x] Make `get_p2p_routing_table` faster

### DIFF
--- a/README.md
+++ b/README.md
@@ -103,27 +103,45 @@ The tests can now be run using:
 
     py.test
 
-Some tests require a connected, (ideally non-booted) SpiNNaker system. To
-run these use:
+This runs a subset of the full test suite which does not require any additional
+hardware.
 
-    py.test --spinnaker HOSTNAME WIDTH HEIGHT
+### Running tests against real hardware
+
+Some tests require a connected, SpiNNaker system. To run these use:
+
+    py.test --spinnaker SPINN_HOSTNAME WIDTH HEIGHT
+
+Though the vast majority of tests will run against any SpiNNaker system with at
+least 4x4 working chips, the complete test suite should run against a single
+SpiNN-5 board. To enable these additional tests, add the `--spinn5` argument:
+
+    py.test --spinnaker SPINN_HOSTNAME 8 8 --spinn5
 
 Other tests require a connected BMP (Board Management Processor, part of
 SpiNN-5 boards). To run these use:
 
-    py.test --bmp HOSTNAME
+    py.test --bmp BMP_HOSTNAME
 
 When a BMP is connected, the test suite will attempt to power-cycle the
 attached board. When a SpiNNaker system is connected, the test suite will
-attempt to boot the system. To skip the power-cycling or booting tests simply
-add the `--no-boot` argument.
+attempt to boot the system.
 
-The test suite ideally should be run against a single SpiNN-5 board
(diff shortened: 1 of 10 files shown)


README.md
```diff
@@ -103,27 +103,47 @@ The tests can now be run using:
...
 
-Some tests require a connected, (ideally non-booted) SpiNNaker system. To
-run these use:
+This runs a subset of the full test suite which does not require any additional
+hardware.
 
-    py.test --spinnaker HOSTNAME WIDTH HEIGHT
+### Running tests against real hardware
+
+Some tests require a connected, SpiNNaker system. To run these use:
```
mundya: Comma unnecessary.

mundya:
Modulo finding a "permanent" home for the proxy this LGTM.

Who Rigged Rodger Rabbit?

mossblaser:
Ugh... I get the feeling that getting a reliable boot at a distance is going to be an issue... :(

Jonathan

mundya:
~~LGTM~~ :(

mossblaser:
Will briefly look into TCP tunneling...

mossblaser:
OK, with any luck once Travis' backlog clears a little these tests should all run perfectly. CSPC276 will become the home for the proxy server and I'm in the process of setting this up.

README.md
```diff
@@ -103,27 +103,55 @@ The tests can now be run using:
...
-add the `--spinn5` argument to enable some additional tests.
+### Running tests against remote hardware
 
-    py.test --spinnaker HOSTNAME 8 8 --spinn5
+Booting a SpiNNaker board requires the (reliable) sending of packets to UDP
+port 54321 which is frequently blocked by ISPs and is not reliable (since UDP
+gives no guarantees, especially on the open internet). As a result, a proxy
+server must be used to communicate with the board. A utility such as
+[`spinnaker_proxy`](https://github.com/project-rig/spinnaker_proxy) can be used
+along side the test suite as follows:
```
mundya: alongside (no space)

mossblaser:
So, it turns out `get_p2p_routing_table` is slow internationally... With any luck I should be able to modify it so that it only reads enough routing table to cover all chips in the system geometry (rather than the whole table).

Since at present about 60 seconds per test are being burned on this (it is run twice, taking 30s each), I think this fix is probably necessary as part of this PR.

mundya:
> Since at present about 60 seconds per test are being burned on this (it is run twice, taking 30s each), I think this fix is probably necessary as part of this PR.

I'd agree I'm afraid.

mossblaser:
Travis is being bizarre but once it runs, it should now take just over 1 minute to run the test suite when it finally runs. Now takes 10 seconds for get_machine tests vs 60 seconds before.

mossblaser:
Average remote test run time now about 1:20 to 1:30, more like it used to be! As a bonus, get_machine should be faster for everyone!

I'd consider this ready when you are, @mundya :)

mundya:
Only grabbing as much p2p table as there is seems sensible.

Rigged

PR #61, opened 2015-03-12
Remove constraint on data length.

### LINKED ISSUE #56: SDPPacket.data capped at 256 bytes
https://github.com/project-rig/rig/blob/master/rig/machine_control/packets.py#L83

mundya:
- Allows data segments of SCP/SDP packets to be unbounded.

Fixes #56.


mossblaser:
Rigged

PR #64, opened 2015-03-15
Hit 100% Statement & Branch Test Coverage

mossblaser:
* A few additional tests have been added to test run various corner cases.
* Note in one case I've pragma'd out some lines in boot: there is no meaningful
  way to test the boot method with non-default arguments, I've pragma
  no-branch'd the handler conditions for these.
* This commit also removes the odd piece of dead-code in internal methods.
* Struct field reads have their endianness explicitly stated.
* The boot method now accepts structs rather than the raw struct data
  since this seemed like a sensible thing to do.

The large number of `pragma: no branch`es in test code asserts is due to a bug in pytest/pytest-cov. When pytest assertion rewriting is used the coverage checker frustratingly will sometimes shout about branch coverage on these assertions. Even better, since the rewriter's internal behaviour is not documented you can't simply tell pytest-cov to ignore asserts (since there won't be any asserts once the rewriting is complete!). Grrr!

### DIFF
--- a/rig/bitfield.py
+++ b/rig/bitfield.py
@@ -497,20 +497,11 @@ class BitField(object):
                    in field.conditions.items()):
                 yield (identifier, field)
 
-    def _potential_fields(self, field_values=None):
+    def _potential_fields(self):
         """Generator of (identifier, field) tuples iterating over every field
         which could potentially be defined given the currently specified field
         values.
-
-        Parameters
-        ----------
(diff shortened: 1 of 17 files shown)


mossblaser:
Would it be fair post this commit to make coveralls shout if coverage drops below 100%?

mundya:
I think that would be fair.

PR #69, opened 2015-03-16
Add Jonathan to setup.py

mossblaser:
Pure vanity, I know, but it is probably fair ;). Awaiting your merge, @mundya


mundya:
I think you could reorder if you want, alphabetic by surname at least :) Otherwise LGTM

Issue #68, opened 2015-03-16
"Recursively allocating memory" / "Grabbing blocks of memory"


mundya:
The `MachineController` provides `sdram_alloc_as_io` as a way of allocating a block of memory and then getting a file-like view of that block of memory.  Sadly, because the tag ~~table~~ range is quite small we're generally going to be limited to performing 1 alloc per core and then laying out that larger block of memory much as we currently do (with something like a pointer table at the start).  I've been trying to think of ways of making this nicer to work with from the host and one thing I was considering was extending `MemoryIO` so that it has a method which means something like "give me the next _n_ bytes of memory as a new `MemoryIO` and move the current pointer to the end of that block".

E.g.:
```python
>>> mem = cn.sdram_alloc_as_io(1024, tag=1)  # I now have 1kB of SDRAM as a file-like
>>> mem1 = mem.alloc_next(400)  # `mem1` is a 400 byte block = mem[0:400]
>>> mem.tell()
400
```

The result is that building an application pointer table can occur alongside getting a different file-like view for each region in the allocated region of SDRAM.

```python
region_lengths = [100, 300, 400, 500]
mem_overall = cn.sdram_alloc_as_io(sum(region_lengths) + len(region_lengths)*4 + APP_PTR_HEADER_LENGTH)
mem_app_ptr = mem_overall.alloc_next(APP_PTR_HEADER_LENGTH + 4*len(region_lengths))
mem_app_ptr.seek(APP_PTR_HEADER_LENGTH)

regions = []
for length in region_lengths:
    mem_app_ptr.write(mem_overall.tell())
    regions.append(mem_overall.alloc_next(length))
```

Some questions:
 - Does this seem like a good idea?
 - Can we think of a better name than `alloc_next` or `alloc_next_as_io`?

mossblaser:
I like the idea, however this seems like a perfect opportunity to use the (rig) Region abstraction. Make a region which accepts a set of regions and it generates the lookup table at the start and can be dumped into a single malloc'd block. `ContainerRegion` perhaps?

mundya:
True, I'm definitely more trying to think of a nice way of getting a smaller block of memory out of a `MemoryIO`. How does this seem for that?

mossblaser:
I'm not sure I understand the specific motivation? Since the application need only grab a pointer to the pointer table (requring one tag per core), couldn't further allocations can be made by native sdram_alloc_as_io calls (without tags)?

mundya:
[...]

PR #72, opened 2015-03-17
Improve `MemoryIO`

mundya:
- Improves the `seek` docstring to mention `os.SEEK_CUR` and friends.
 - If `read` is called with no (or negative) argument then it is interpreted as being a read until the end of the memory region.
 - Implements the suggestions from #68 
 - flake8

### DIFF
--- a/docs/source/control.rst
+++ b/docs/source/control.rst
@@ -168,7 +168,7 @@ into applications and also ensures reads and writes are constrained only to the
 allocated region. For example::
 
     >>> # Allocate 1024 bytes of SDRAM with tag '3' on chip (0, 0)
-    >>> block = mc.sdram_alloc_as_io(1024, 3, 0, 0)
+    >>> block = mc.sdram_alloc_as_filelike(1024, 3, 0, 0)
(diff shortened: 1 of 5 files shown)


mossblaser:
`alloc_next_as_io` -> `create_view` or `create_slice`?

Perhaps?

Aside from that, that looks fair, I think!

mundya:
I like `create_view`... I'm not so good at naming things.

(Though better than some, tennis was originally called ["Sphairistiké"](http://www.britannica.com/EBchecked/topic/587387/tennis))

mundya:
Hold off on this, I'd like to try something else.

mundya:
I'd like to merge this soonish, I've removed `create_view` until we can decide better semantics for it.  I've also renamed `MachineControl.sdram_alloc_as_io` to `sdram_alloc_as_filelike`.  I'm currently reading up on the correct way to present memoryviews.

mossblaser:
Fair enough. I've updated the docs. Assuming travis doesn't barf I'll merge that in ASAP.

Issue #63, opened 2015-03-14
Future Improved IO Plans/Thoughts


mossblaser:
* Blocking interfaces should always be present and should continue to work just like they do now
* Plenty of performance is lost to waiting for roundtrips on SCP commands
  * Must ask ST what restrictions apply to current implementations/his thoughts on having multiple open SCP requests
  * In many cases, blocking functions make lots of potentially-parallel SCP calls (e.g. get_machine), thus many functions with blocking interfaces would end up being faster for free.
* In the future, opportunities will exist for using multiple endpoints simultaneously for SCP
  * Should ask ST about side-effects of communicating with the system from multiple ethernet (or otherwise) links. I would guess none.
  * It might not be necessary/beneficial to limit comms to just use of the local ethernet connection
* Alternative interface should allow the queuing/dispatching of many SCP commands at once
  * Should, behind the scenes, schedule these commands appropriately (i.e. limit maximum outstanding commands)
  * Some plan needs to be implemented to handle deferred responses
    * Callbacks on completion (messy)
    * ~~Use a "real" framework for this (e.g. Twisted)~~
    * Methods could be "two-phase" generators
      * Calling the method yields after the command is either queued or sent
      * Doing a next() on the generator would then produce the received response, blocking until the command has both been sent and the response received. The generator could then yield the response or throw an appropriate exception (e.g. timeout)
      * Potentially, enabling this mode could simply be an argument of all the existing commands -- better still, it would make it possible to use a decorator for this function. Also, as an argument it can be set by the context -- schweet.
      * Additionally provide some flush-style function which blocks until all commands have been sent and responded to.
      * Clearly JH has taken a liking to this idea

mundya:
> - Use a "real" framework for this (e.g. Twisted)

Twisted would require rethinking most of the API, for the record I'm against using it.

mossblaser:
Just had a chat with ST:
* There is a hard-limit on the number of outstanding connections going through a ethernet-connected monitor imposed by the transient IPTag pool size (a transient IP tag is created for every SCP cmd). This can be queried from the machine. This can also be cheaply extended if it proves valuable.
* If the bandwidth/latency ratio for loading a core is sufficiently good (unverified but highly plausible) then the only situations where parallel requests are useful is when many cores are being loaded/polled simultaneously.
* Having multiple outstanding SCP requests where each request is to a different core doesn't require any new flow-control mechanisms in SCP.
* ST recommends only using the ethernet connection local to a given chip for comms.

[...]

PR #73, opened 2015-03-19
Asynchronous, parallel IO

mossblaser:
Attempts to implement something along the lines of #63.

* [x] Non-blocking, parallel SCP implementation (SCPProtocol)
* [x] Remove old SCPConnection and port to SCPProtocol (+ asyncio/trollius)
* [x] Non-blocking MachineController and BMPController
  * [x] BMPController
  * [x] MachineController
* [x] Non blocking boot function(?) This would be a ~~certain amount of effort~~ easy to do for dubious gain...
* [ ] Non blocking MemoryIO...
* [x] Blocking wrapper for most functions
* [ ] Discover number of IPTags + set small IPTag timeout
* [ ] Benchmarking (will need to get the new SARK/SCAMP for this)
  * [x] Old approach v.s. this approach limited to a single connection (i.e. implementation overhead, or not, hopefully)
  * [x] Old approach v.s. parallel requests
  * [x] Varying number of parallel requests
  * [x] Single vs multiple Ethernets
  * [ ] This v.s. pacman (sanity check - sometimes python threads are supa-fast)
  * [ ] This benchmark v.s. ST's own measurements
* [x] Multiple Ethernets
  * [x] Multiple Ethernet connection auto-discovery
  * [x] Auto-select right connection to use
  * [x] Test against the 105 machine or something
* [ ] Documentation
  * [ ] Updates for "basic, everyday use" (mostly just "this is fast, enjoy!" and "call this to use **moar** ethernet connections)
  * [ ] Tutorial on using the parallel stuff for advanced users


mundya:
Parallel good, does this do sliding window?

mossblaser:
But of course :)

mossblaser:
I should clarify that by parallel (at this point in development) I really mean that many requests can be outstanding in parallel :). Most of the hard work there should be done since all the scheduling logic is in the SCPProtocol class, everything else should be a case of porting the rest which should be reasonably straight-forward.

mundya:
> many requests can be outstanding in parallel :)

Ah, ok, sending to many monitor processes in parallel is trivial anyway. Awesome.

rig/machine_control/scp_protocol.py
```diff
@@ -0,0 +1,411 @@
...
+                # The CMD_RC indicates that the command failed. Raise an
+                # appropriate exception detailing the original packet sent.
+                orig_packet = SCPPacket.from_bytestring(
+                    out_req.queued_request.data[2:])
+                if not out_req.queued_request.future.cancelled():
+                    out_req.queued_request.future.set_exception(
+                        self._error_codes[packet.cmd_rc](
+                            "Error in response to packet with arguments: "
+                            "cmd_rc={}, arg1={}, arg2={}, arg3={}; "
+                            "sent to core ({},{},{}).".format(
```
mundya: For reference, having implemented this I don't like it. I'd rather pass the orig_packet and the received failure and let the exception decide how to best display the error...

mossblaser: Sounds like a sane idea to me :). (will try and remember to do this!).

mossblaser:
@mundya Any ideas how I should deal with MemoryIO in this new non-blocking world?

mundya:
Ooh, unless this is blocking you can I have a go at it to test my usage of the api?

mossblaser:
Sure, you'd be more than welcome :) just await the Machine Controller
changes being applied, if you would :)

mundya:
Coincidentally, am I correct in thinking that if we switched all methods in `MachineController` and `BMPController` to use `ContextMixin.use_named_contextual_arguments` and re-ordered the decorators as:

```python
@ContextMixin.use_named_contextual_arguments(x=Required, y=Required, p=0, async=False)
@LookBlockingMixin.look_blocking
@trollius.coroutine
def command(arg1, arg2, **kwargs):
    x = kwargs.pop("x")
    y = kwargs.pop("y")
    p = kwargs.pop("p")
    # do stuff
```

We would get the ability to do:
```python
with cn(async=True):
    f1 = cn.do_something()
    f2 = cn.do_something_else()

cn.loop.run_until_complete(trollius.gather(f1, f2), loop=cn.loop)  # or similar
```

Or, with a further minor tweak:
```python
cn.update_current_context(asycn=True)

# Now all commands run with cn are asynchronous by default
```

mossblaser:
You are correct in this thinking, however I would much prefer to fix the look_blocking mixin to modify the function metadata so that the contextual argument stuff spots that it has an async argument in the first place (so you don't need to mention it everywhere!). This is informally on the list. I'd also like it if the use named contextual arguments could update the function signature too for auto-documentation purposes.

mundya:
> I'd also like it if the use named contextual arguments could update the function signature too for auto-documentation purposes.

AFAIK this is not at all trivial, though I am looking into it.

mossblaser:
Just performed a very approximate experiment on the performance of this branch
vs the old one with various functions and numbers of allowed outstanding
commands and here are the results:

            |      get_machine()      |   read(1mb, x=0, y=0)   |   read(1mb, x=1, y=1)
    name    |  min     mean     std   |  min     mean     std   |  min     mean     std
    ------- | ------  -------  ------ | ------  -------  ------ | ------  -------  ------
    old     | 0.0469  0.0502   0.0152 | 1.8425  1.8936   0.2755 | 3.2515  3.3699   0.6202
    new_1   | 0.0719  0.0751   0.0260 | 2.0439  2.0858   0.0849 | 3.4833  3.8421   2.8687
    new_2   | 0.0639  0.0660   0.0239 | 1.4825  1.4929   0.0227 | 2.6938  3.3934   1.9673
    new_4   | 0.0616  0.0632   0.0242 | 1.2138  1.2250   0.0297 | 2.2166  2.9085   1.3611
    new_8   | 0.0621  0.0641   0.0257 | 1.2184  1.2296   0.0284 | 2.1587  2.8955   1.7335

In the table old is `master`, `new_X` is `nonblocking_io` with the SCPProtocol's
`max_outstanding` value set to `X`. `get_machine()` was run 100 times, the
`read()`s 10 times.

This experiment was performed on my machine against a SpiNN-5 board connected
to the same switch as my machine and running SC&MP v1.33 (i.e. not the new
version). Network latency reported by `ping`: around 0.3 ms.

Not measured formally but it was noted that the new implementation will peg the
CPU while the old implementation will generally sit closer to 40% usage most of
the way through the benchmark.

Takeaways:

* An effort must be made to address the computational overheads involved
  here -- this could quickly limit scaling.
* The overhead involved in running all the coroutine magic is clearly making
  itself known for `get_machine()` where lots of small commands are involved.
  Certainly, the GC is going to be getting plenty of thrashing and lots of
  (Python) context switching will be occurring.
  * In the presence of higher network latency the increased window may/may-not
    have a greater positive impact to recover some of this overhead.
  * The new version of SC&MP's improved network performance may/may-not help
    recover some of this.
* With a single outstanding command, reads are slower than they used to be, with
  multiple outstanding commands the performance quickly starts beating the old
  implementation.
  * Given I expect most applications will spend much of their I/O doing bulk
    transfer style operations, this probably means even if things don't improve
    for small operations with the new SC&MP, this tradeoff is profitable.
  * My hope is that the new SC&MP will make even better use of multiple
    outstanding connections arriving.

TODOs:

* Repeat these experiments over a higher-latency connection. (Hypothesis:
  greater benefits for `get_machine()` and (maybe) `read()` on new version with
  outstanding commands.)
* Attempt these experiments with a multi-board system with multiple Ethernet
  connections. (Hypothesis: improved `get_machine()` if the overhead isn't
  awful, linear scaling of `read()` if I run two reads on different boards in
  parallel).
* Run a python profiler against this to see where overheads seem to be

mossblaser:
Here are some quick results from running the new library against multiple
Ethernet connections to the three-board system on my desk (via a 100 MB/s
switch...).

                       |     single Ethernet     |    three Ethernets
    conns  function    | min     mean     std    | min     mean     std
    -----  ----------- | ------  ------   ------ | ------  ------   ------
    1      get_machine | 0.2000  0.2070   0.0533 | 0.1733  0.1814   0.0746
    1      read        | 1.9979  2.0114   0.0310 | 2.0435  2.0631   0.0438
    1      read_three  | 9.8571  10.1961  0.5131 | 3.9689  4.3103   0.6640
    -----  ----------- | ------  ------   ------ | ------  ------   ------
    2      get_machine | 0.1757  0.1870   0.1052 | 0.1681  0.2011   0.3029
    2      read        | 1.4774  1.5269   0.2861 | 1.5197  1.5754   0.3144
    2      read_three  | 5.5351  5.6314   0.1484 | 3.7278  3.7741   0.0983
    -----  ----------- | ------  ------   ------ | ------  ------   ------
    4      get_machine | 0.1728  0.1797   0.0457 | 0.1652  0.1763   0.1514
    4      read        | 1.2386  1.2595   0.0528 | 1.2982  1.3550   0.1766
    4      read_three  | 4.2686  4.7567   1.2995 | 3.7782  3.8079   0.0825
    -----  ----------- | ------  ------   ------ | ------  ------   ------
    8      get_machine | 0.1705  0.1767   0.0442 | 0.1634  0.2009   2.9492
    8      read        | 1.2164  1.2205   0.0114 | 1.2278  1.2390   0.0365
    8      read_three  | 5.2510  5.5308   0.4746 | 3.6458  3.6642   0.0264


Here `read_three` reads 1 MB from the three Ethernet connected chips
simultaneously.

Use of multiple connections has been verified in as much as the relevant lights
blink only after connection discovery.

Observations:
* Performance scaling of `read_three` occurs but only approaches linear scaling
  with a single connection. Sadly, with larger numbers of connections there is
  still an improvement but this is limited.
* CPU usage maxed out the whole time

Sadly I think this means that scaling isn't really going to be a thing with
`asyncio`, at least not with the current implementation...

mossblaser:
Regrettably I'm mothballing this implementation on account of its lacklustre performance. Follow #77 for its successor!

PR #74, opened 2015-03-21
Add `update_current_context` to `ContextMixin`

mundya:
- `update_current_context` allows changing of a context without
   entering a `with` block.
 - Corrects `Context.__init__` to copy the dictionary it is passed, not
   doing so caused some bugs when introducing `update_current_context`.

Expected use case:
```python
cn = MachineController(...)
cn.update_current_context(async=True)  # Retains the app_id from the default context.
```

This allows the final example in my comment on #73.

### DIFF
--- a/docs/source/control.rst
+++ b/docs/source/control.rst
@@ -186,6 +186,12 @@ reduce the repetition, Python's ``with`` statement can be used::
     ...     block_addr = mc.sdram_alloc(1024, 3)
     ...     mc.write(block_addr, b"Hello, world!")
 
+Alternatively, the current context can be modified by calling
+:py:meth:`~.MachineController.update_current_context`::
+
+    >>> # Following this call all commands will use app_id=56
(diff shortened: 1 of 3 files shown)


mossblaser:
A brief mention in https://github.com/project-rig/rig/blob/master/docs/source/control.rst#context-managers-1 wouldn't go amiss :). Pending that, LGTM :).

mundya:
Uh, is there an easy way to only add the docs for *one* inherited method?

```
.. autoclass:
    :members:
    :special-members:
```
Doesn't include `update_current_context` but:

```
.. autoclass:
    :members:
    :special-members:
    :inherited-members:
```
Includes _everything_ and `:inherited-members: update_current_context` doesn't limit it either...  Is this not worth worrying about provided there's some comment in Context Manager section?

mossblaser:
Alas this has been a source of much pain as far as I know. I think for now that mention in the tutorial will do...

PR #83, opened 2015-04-01
Optimise code for Python interpreter performance.

mossblaser:
Since Rig has now grown sufficiently that the Python interpreter spends almost
1ms running its parser on the codebase, it has become obvious that in order to
maintain reasonable performance some effort should be made to reduce the input
size.

My own benchmarking has found these changes to make an almost-measurable
difference. Further, by compressing the code, more can be seen on the screen at
any one time making it far quicker to maintain and debug.


mossblaser:
Regarding the slight code coverage reduction, this was a necessary trade-off since the inclusion of many `# pragma: no cover` lines proved especially wasteful. I'm sure you'll agree that the vast reduction in code size and, because number of bugs is proportional to number of lines, the number of potential bugs, this change should be considered positive overall.

mundya:
Sorry, I'm not convinced that unreadable/massively ugly code is a price worth paying to reduce a compile stage that for most users will occur once only.

mundya:
I'm a fool.  Good,

PR #89, opened 2015-04-10
Ensure Bitfield.__call__ uses `type`.

mundya:
Allows subclassing of `BitField`, inasmuch as `__call__` creates a new instance of the same class rather than always calling `BitField`.

Previously:
```python
>>> class MyBitField(BitField):
...     pass
>>>
>>> x = MyBitField()
>>> x.add_field("spam")
>>> type(x(spam=3))
rig.bitfield.BitField
```

Whereas now:
```python
>>> type(x(spam=3))
__main__.MyBitField
```


mundya:
Merged into #91.

Issue #90, opened 2015-04-10
Mixing positioned and unpositioned fields in BitField doesn't work as expected


mundya:
The BitField docs state: "It is perfectly valid to mix fields both with and without allocated lengths and positions." as it's possible to assign lengths and starting positions independently my assumption would be that I can do the following:

```python
>>> from rig.bitfield import BitField
>>> bf = BitField()
>>> bf.add_field("not_positioned")
>>> bf.add_field("positioned_not_sized", start_at=0)
```

Should I be able to do this?

If I then attempt to use this everything goes awry:
```python
>>> x = bf(not_positioned=255, positioned_not_sized=256)
>>> x.assign_fields()
>>> hex(x.get_mask(field='positioned_not_sized'))
'0xff'  # This is what I'd expect
>>> hex(x.get_mask(field='not_positioned'))
'0x1ff'  # This definitely is not (expecting 0x1ff00)
```

If I shouldn't be trying to do this then the documentation needs clarifying; otherwise there's a bug...  I've run this test against the version of BitField in `master`.

mundya:
Glancing through the tests it appears that fields grow from MSB to LSB; is this right?  If so, is there a way I can force fields to be LSB-aligned?  Also, if so, surely my example should throw some kind of exception rather than just positioning fields over each other?

PR #91, opened 2015-04-10
Fix BitField allocation of fixed-position fields

mundya:
Ensures that fields with a fixed starting position but no fixed length are correctly included in `assigned_bits` during the process of assigning fields.

This highlights/causes immense problems with using breadth-first search to assign field lengths; @mossblaser, do you have any thoughts on this?
My view would be that the fix comes in two flavours:
 - Make it possible to allow fields to be allocated from MSB to LSB or vice versa and ensure that one only positions variable length fields at the opposite end (rather like keeping the stack and the heap separate).  I'll tack this onto this PR shortly.
 - Have a much more complex field assignment procedure which works from the leaves back to the root (and may have to rip up and retry at various points).

### DIFF
--- a/rig/bitfield.py
+++ b/rig/bitfield.py
@@ -6,7 +6,7 @@ hierarchical bit-fields.
 See the :py:class:`.BitField` class.
 """
 
-from collections import OrderedDict
+from collections import namedtuple, OrderedDict
 
 from math import log
 
@@ -142,14 +142,15 @@ class BitField(object):
                 if other_field.start_at is not None:
                     other_start_at = other_field.start_at
                     other_end_at = other_start_at + (other_field.length or 1)
+
                     if end_at > other_start_at and other_end_at > start_at:
-                            raise ValueError(
-                                "Field '{}' (range {}-{}) "
-                                "overlaps field '{}' (range {}-{})".format(
-                                    identifier,
-                                    start_at, end_at,
-                                    other_identifier,
-                                    other_start_at, other_end_at))
+                        raise ValueError(
+                            "Field '{}' (range {}-{}) "
+                            "overlaps field '{}' (range {}-{})".format(
+                                identifier,
+                                start_at, end_at,
+                                other_identifier,
+                                other_start_at, other_end_at))
 
         # Normalise tags type
         if type(tags) is str:
@@ -221,7 +222,7 @@ class BitField(object):
             self.fields[identifier].max_value = max(
                 self.fields[identifier].max_value, value)
 
-        return BitField(self.length, self.fields, field_values)
+        return type(self)(self.length, self.fields, field_values)
 
     def __getattr__(self, identifier):
         """Get the value of a field.
@@ -315,7 +316,9 @@ class BitField(object):
             `assign_fields()` has not been called when a field's size/position
             has not been fixed.
         """
-        assert not (tag is not None and field is not None)
+        if tag is not None and field is not None:
+            raise TypeError("get_mask() takes exactly one keyword argument, "
+                            "either 'field' or 'tag' (both given)")
 
         # Build a filtered list of fields to be used in the mask
         if field is not None:
@@ -361,19 +364,55 @@ class BitField(object):
         Users should typically call this method after all field values have
         been assigned, otherwise fields may be fixed at an inadequate size.
         """
-        # We must fix fields at every level of the heirarchy sepeartely
+        # We must fix fields at every level of the hierarchy separately
         # (otherwise fields of children won't be allowed to overlap). Here we
-        # do a breadth-first iteration over the heirarchy, fixing the fields at
-        # each level.
-        unsearched_heirarchy = [BitField(self.length, self.fields)]
-        while unsearched_heirarchy:
-            ks = unsearched_heirarchy.pop(0)
-            ks._assign_enabled_fields()
-            # Look for potential children in the herarchy
+        # do a breadth-first iteration over the hierarchy to fix fields with
+        # given starting positions; then we do depth-first to fix other fields.
+        #
+        # The breadth-first ensures that children's fixed position fields must
+        # fit around the fixed position fields of their parents.  The depth
+        # first search for variable position fields ensures that parents don't
+        # allocate fields in positions which would collide with fixed and
+        # variable position fields their children have already allocated.
+
+        class Node(namedtuple("Node", "bitfield children")):
+            """Node used in depth-first search of bitfields."""
+            def recurse_assign_unfixed_fields(self, assigned_bits=0,
+                                              excluded_fields=set()):
+                # Call this for the children first
+                excludes = set(i for (i, f) in self.bitfield._enabled_fields())
+                new_assigned_bits = assigned_bits
+                for child in self.children:
+                    # Children can assign bits independently
+                    new_assigned_bits |= child.recurse_assign_unfixed_fields(
+                        assigned_bits, excludes)
+
+                # Now assign locally
+                assigned_bits =\
+                    self.bitfield._assign_enabled_fields_without_fixed_start(
+                        new_assigned_bits, excluded_fields
+                    )
+
+                return assigned_bits
+
+        root = Node(BitField(self.length, self.fields), list())
+        unsearched_hierarchy = [root]
+
+        # Build a tree of Node describing the bitfield hierarchy and assign all
+        # fields with a fixed starting position.
+        while unsearched_hierarchy:
+            parent = unsearched_hierarchy.pop(0)
+            ks = parent.bitfield
+
+            # Assign all fields with a fixed starting position
+            ks._assign_enabled_fields_with_fixed_start()
+
+            # Look for potential children in the hierarchy
             for identifier, field in ks._potential_fields():
                 enabled_field_idents = set(i for (i, f) in
                                            ks._enabled_fields())
                 set_fields = {}
+
                 for cond_ident, cond_value in field.conditions.items():
                     # Fail if not a child
                     if cond_ident not in enabled_field_idents:
@@ -382,8 +421,14 @@ class BitField(object):
                     # Accumulate fields which must be set
                     if getattr(ks, cond_ident) is None:
                         set_fields[cond_ident] = cond_value
+
                 if set_fields:
-                    unsearched_heirarchy.append(ks(**set_fields))
+                    child = Node(ks(**set_fields), list())
+                    parent.children.append(child)
+                    unsearched_hierarchy.append(child)
+
+        # Assign all fields with variables starting positions
+        root.recurse_assign_unfixed_fields()
 
     def __eq__(self, other):
         """Test that this :py:class:`.BitField` is equivalent to another.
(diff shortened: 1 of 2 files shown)


mundya:
It's worth noting that this partially fixes my problem from #90 but inserts a hideous new issue:

```python
>>> from rig.bitfield import BitField

>>> ks = BitField(32)
>>> ks.add_field("user")

>>> ks0 = ks(user=1)
>>> ks0.add_field("spam", start_at=0)

# At this point I would like the user field to be pushed towards the MSB by the
# requirement that "spam" is fixed at the LSB.  Instead "user" is assigned to
# start at the LSB which then obviously causes problems when trying to assign
# "spam".
>>> ks.assign_fields()
ValueError: None-bit field 'spam' with fixed position does not fit in bit field.
```

One fix for this might be:
```python
>>> ks = BitField(32)
>>> ks.add_field("user", grow_from=ks.msb)  # ks.msb == 31
>>> ks0 = ks(user=1)
>>> ks0.add_field("spam", start_at=0)
>>> ks.assign_fields()
>>> hex(ks.get_mask(field='user'))
'0x80000000'
>>> hex(ks0.get_mask(field='spam'))
'0x00000001'
```

mundya:
That last example is now:

```python


>>> ks = BitField(32)
>>> ks.add_field("user", relative_to_lsb=False)
>>> ks0 = ks(user=1)
>>> ks0.add_field("spam", start_at=0)
>>> ks.assign_fields()
>>> hex(ks.get_mask(field='user'))
'0x80000000'
>>> hex(ks0.get_mask(field='spam'))
'0x00000001'
```

mundya:
@mossblaser I'm not sure how to deal with the failing test: It's a test that I believe should fail (i.e., a `ValueError` should be raised when you try to do what the test does) unless we decide to rewrite the assignment code to do some form of priority ordering for fixed start positions.

mundya:
After some further thinking I believe an alternative assignment algorithm is to use breadth-first search to assign lengths to all fields which have a start position specified and to then use depth-first search to apply (from the leaves upward) a start position and length to fields which have neither specified.  As a result all hard "this field should be here" constraints should be met before fitting in additional fields where possible.

E.g.
```python
>>> ks = BitField(32)
>>> ks.add_field("user")

>>> ks0 = ks(user=1)
>>> ks0.add_field("spam", start_at=0)
>>> ks0(spam=0xff)

>>> ks.assign_fields()
```
Should result in:
```python
>>> hex(ks0.get_mask(field="spam"))
'0xff'
>>> hex(ks0.get_mask(field="user"))
'0x100'
```

I'm currently implementing that, but will probably finish that off for tomorrow.

mundya:
Pulling everything together:

The original problem I noticed was that the commands:
```python
>>> from rig.bitfield import BitField
>>> ks = BitField(32)
>>> ks.add_field("f1", length=4)
>>> ks_0 = ks(f1=0)
>>> ks_0.add_field("f2", start_at=0, length=4)
>>> ks.assign_fields()
```
Resulted in impossible field assignment; viz:
```python
>>> hex(ks.get_mask(field="f1"))
'0xf'
>>> hex(ks_0.get_mask(field="f2"))
'0xf'
```
Obviously bad.

The fix for this was to ensure that fields with fixed start positions were included in the mask of bits which had been assigned to fields (this counts as a bugfix and should be retained even if this PR doesn't make it!).

This highlighted some nastiness in that `f1` would be assigned bits `0xf` before `f2` was processed - meaning that `f2` could never be assigned.

Fixing this requires that all fields with fixed start positions are sized and positioned _before_ fields which can be arbitrarily moved.  My solution to this is to perform two passes over the trees of bitfields: a breadth-first (as before) to assign fields with fixed position and a depth-first to assign other fields.  Sadly, this completely breaks @mossblaser's nice field ordering.

As an example of this working:
```python
>>> ks = BitField(32)
>>> ks.add_field("split", length=1)
>>> ks_0 = ks(split=0)
>>> ks_0.add_field("s0", length=4, start_at=0)
>>> ks_1 = ks(split=1)
>>> ks_1.add_field("s1", length=8, start_at=0)
>>> ks.assign_fields()
>>> hex(ks.get_mask(field="split"))
'0x100'
>>> hex(ks_0.get_mask(field="s0"))
'0xf'
>>> hex(ks_1.get_mask(field="s1"))
'0xff'
```

The tree for depth-first traversal is constructed on calling `add_field`; consequently some nasties are:
```python
>>> ks = BitField(32)
>>> ks.add_field("a")
>>> ks.add_field("b")
>>> ks_0 = ks(a=0, b=0)  # This is a child of ks, not of ks(a=0)(b=0)...
```
I suspect there is a nicer way to implement this that will avoid this occurring but I currently can't think of it...

I've also added an option to allow fields to be specified in terms of their position from either the LSB (default) or the MSB.

This is a pretty big change from what we had before so thoughts from @mossblaser (and @neworderofjamie) are definitely sought! :smiley:

rig/bitfield.py
```diff
@@ -137,12 +148,27 @@ def add_field(self, identifier, length=None, start_at=None, tags=None):
...
-                    other_end_at = other_start_at + (other_field.length or 1)
-                    if end_at > other_start_at and other_end_at > start_at:
+                    if other_field.relative_to_lsb:
+                        other_start_at = other_field.start_at
+                        other_end_at = \
+                            other_start_at + (other_field.length or 1)
+                    else:
+                        other_start_at = (self.length - other_field.start_at -
+                                          (other_field.length or 1))
+                        other_end_at = (self.length - other_field.start_at)
```
mundya: Hmm... this will break if the field we're comparing against has already been assigned.  When assigning a field the "relative_to_lsb" tag should be set to true.

rig/bitfield.py
```diff
@@ -536,43 +603,99 @@ def _potential_fields(self):
...
+                        field.length, identifier)
+                )
+
+            # Mark these bits as assigned
+            assigned_bits |= field_bits
+
+        # Check that the calculated field is within the bit field
+        if start_at + length <= self.length:
+            field.length = length
+            field.start_at = start_at
```
mundya: `field.relative_to_lsb = True`

rig/bitfield.py
```diff
@@ -536,43 +609,100 @@ def _potential_fields(self):
...
+        fields which do not have one.
+
+        Parameters
+        ----------
+        exclude_fields : set
+            Exclude fields which will be assigned later.
+        """
+        unassigned_fields = list()
+        for identifier, field in self._enabled_fields():
+            if field.length is None or field.start_at is None:
```
mundya: For safety sake I guess this should be `and`

rig/bitfield.py
```diff
@@ -536,43 +572,90 @@ def _potential_fields(self):
             else:
                 blocked.add(identifier)
 
-    def _assign_enabled_fields(self):
-        """For internal use only. Assign a position & length to any enabled
-        fields which do not have one.
+    def _assign_enabled_fields_with_fixed_start(self):
+        """For internal use only. Assign a length to any enabled fields which
+        do not have one.
```
mossblaser: "but do have a fixed position"

rig/bitfield.py
```diff
@@ -536,43 +572,90 @@ def _potential_fields(self):
...
-        unassigned_fields = []
+        unassigned_fields = list()
         for identifier, field in self._enabled_fields():
             if field.length is not None and field.start_at is not None:
                 assigned_bits |= ((1 << field.length) - 1) << field.start_at
-            else:
+            elif field.start_at is not None:
                 unassigned_fields.append((identifier, field))
 
+        # Sort the unassigned fields in order of fixed start position
```
mossblaser: why?

mundya: No reason.

rig/bitfield.py
```diff
@@ -536,43 +572,90 @@ def _potential_fields(self):
...
+        fields which do not have one.
 
-            # Check that the calculated field is within the bit field
-            if start_at + length <= self.length:
-                field.length = length
-                field.start_at = start_at
+        Parameters
+        ----------
+        exclude_fields : set
+            Exclude fields which will be assigned later.
```
mossblaser: Example?

rig/bitfield.py
```diff
@@ -536,43 +572,90 @@ def _potential_fields(self):
...
-                        start_at = bit
-                        assigned_bits |= field_bits
-                        break
+            assigned_bits = self._assign_enabled_field(
+                assigned_bits, identifier, field)
+
+    def _assign_enabled_fields_without_fixed_start(self, assigned_bits,
+                                                   exclude_fields=set()):
+        """For internal use only. Assign a length and position to any enabled
+        fields which do not have one.
```
mossblaser: "and do not yet have a fixed position."

mundya:
Must squash prior to merge!

rig/bitfield.py
```diff
@@ -536,43 +572,88 @@ def _potential_fields(self):
...
 
-            # Check that the calculated field is within the bit field
-            if start_at + length <= self.length:
-                field.length = length
-                field.start_at = start_at
+        Parameters
+        ----------
+        exclude_fields : {identifier, ...}
+            Exclude fields which will be assigned later, for example those
+            which will be assigned by keyspaces higher up the hierarchy.
```
mossblaser: s/keyspaces/fields/ ?

rig/bitfield.py
```diff
@@ -536,43 +572,88 @@ def _potential_fields(self):
...
+
+        for identifier, field in unassigned_fields:
+            assert identifier not in exclude_fields
+            assigned_bits |= self._assign_enabled_field(
+                assigned_bits, identifier, field)
+
+        return assigned_bits
+
+    def _assign_enabled_field(self, assigned_bits, identifier, field):
+        length = field.length
```
mossblaser: Docstring?

rig/bitfield.py
```diff
@@ -361,15 +364,43 @@ def assign_fields(self):
...
+
+                return assigned_bits
+
+        # We must fix fields at every level of the hierarchy separately
         # (otherwise fields of children won't be allowed to overlap). Here we
-        # do a breadth-first iteration over the heirarchy, fixing the fields at
-        # each level.
-        unsearched_heirarchy = [BitField(self.length, self.fields)]
+        # do a breadth-first iteration over the hierarchy to fix fields with
+        # given starting positions; then we do depth-first to fix other fields.
```
mossblaser: (could you add a brief motivation to "why" each type of iteration in this comment! thanks!)

rig/bitfield.py
```diff
@@ -361,15 +364,49 @@ def assign_fields(self):
...
+        #
+        # The breadth-first ensures that children's fixed position fields must
+        # fit around the fixed position fields of their parents.  The depth
+        # first search for variable position fields ensures that parents don't
+        # allocate fields in positions which would collide with fixed and
+        # variable position fields their children have already allocated.
+        root = Node(BitField(self.length, self.fields), list())
+        unsearched_hierarchy = [root]
+
+        while unsearched_hierarchy:
```
mossblaser: # Build a tree of Nodes describing the dependencies of fields in the BitField.

rig/bitfield.py
```diff
@@ -361,15 +364,49 @@ def assign_fields(self):
...
         # (otherwise fields of children won't be allowed to overlap). Here we
-        # do a breadth-first iteration over the heirarchy, fixing the fields at
-        # each level.
-        unsearched_heirarchy = [BitField(self.length, self.fields)]
-        while unsearched_heirarchy:
-            ks = unsearched_heirarchy.pop(0)
-            ks._assign_enabled_fields()
-            # Look for potential children in the herarchy
+        # do a breadth-first iteration over the hierarchy to fix fields with
+        # given starting positions; then we do depth-first to fix other fields.
```
mossblaser: I'd move the above comment to the top of the method as an overview comment

mundya: Done.  Not sure why this line not survived...

mossblaser: Oh cool: those line refs refer to the actual line not just the line number! Clever github!

rig/bitfield.py
```diff
@@ -361,19 +364,60 @@ def assign_fields(self):
...
+                    new_assigned_bits |= child.recurse_assign_unfixed_fields(
+                        assigned_bits, excludes)
+
+                # Now assign locally
+                assigned_bits =\
+                    self.bitfield._assign_enabled_fields_without_fixed_start(
+                        new_assigned_bits, excluded_fields
+                    )
+
+                # Check that everything HAS been assigned
```
mossblaser: Is there some user-error that can cause this to happen or is this just out of lack of certainty that the algorithm works? (If the former, jolly good, if the latter...)

mundya: Neither, just being defensive...

mossblaser:
Excellent piece of work thar! :D Thanks for all the fixes!

Rigged

PR #92, opened 2015-04-15
Fix Links.from_vector when "spiralling" in 2x2

mossblaser:
Handle the following special case which much trouble was made to handle in other
parts of the process but not here (because I'm stupid).

When we've got a 2xN or Nx2 system (N >= 2) we can "spiral" around the Z axis to
reach places which normally wouldn't be accessible. Below shows the two examples
where we can make a "diagonal hop" in the direction not usually allowed by
"spiralling around" on the Z axis.

    (x+1, 0) <-> (x+0, 1)        (1, y+0) <-> (0, y+1)
              /                        |   |   |
        --+--/+---+--                  +---+---+
          | . |   |                    | . |   |/
        --+---+---+--                  /---+---/
          |   | . |                   /|   | . |
        --+---+/--+--                  +---+---+
              /                        |   |   |

### DIFF
--- a/rig/machine.py
+++ b/rig/machine.py
@@ -112,6 +112,21 @@ _link_direction_lookup = {
 }
 _direction_link_lookup = {l: v for (v, l) in iteritems(_link_direction_lookup)}
 
+# Special case: Lets assume we've got a 2xN or Nx2 system (N >= 2) where we can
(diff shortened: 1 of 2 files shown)


mundya:
Welcome back @mossblaser 

LGTM

PR #93, opened 2015-04-15
Fix incorrect variable name.

mossblaser:


mundya:
LGTM.

mundya:
Do we have a test that checks what happens when this is done?  The line must have been being executed, but were we checking what other things occurred?

mossblaser:
We do not. I've just been trying to think of a sensible test which would catch this bug but am struggling...

mundya:
Probably not worth it... :worried:

mossblaser:
Yeah :/ as much as I like all bug fixes to come with a test, this doesn't seem like a sensible time for that.

Issue #94, opened 2015-04-15
Feature Proposal: BitField Namespacing


mossblaser:
To alleviate the need for project-rig/nengo_spinnaker#1 I propose allowing fields to exist in their own namespace in the hierarchy allowing:

* External users need no longer worry about colliding with fields they cannot see/access
* Multiple external users could define fields with a particular name (agreed on externally) which could then be accessed transparently elsewhere. For example, external devices may have a specific packet format enforced but have essentially the same set of fields as internal packets -- it would be desirable for a user to not need to know which type they're using.
* Would finally strictly enforce the internal data structures into a real tree...

PR #95, opened 2015-04-16
Add namespace support to Bitfields

mossblaser:
As proposed in #94, allows multiple fields in a BitField to share the same name
so long as they are mutually exclusive.

This commit fundamentally changes the underlying datastructure in the BitField
class from a somewhat ad-hoc dictionary-with-dependency-list form into a true
tree format. This change alone is a "good thing" (TM) and has also simplified
implementation in some places.

Breaking changes
------------------------

* Exception types changed in some cases
* `get_tags` now only works when a field is enabled

### DIFF
--- a/rig/bitfield.py
+++ b/rig/bitfield.py
@@ -1,12 +1,14 @@
 """A system for defining and representing bit fields.
 
-A common use-ase for this module is defining SpiNNaker routing keys based on
+A common use-case for this module is defining SpiNNaker routing keys based on
 hierarchical bit-fields.
 
 See the :py:class:`.BitField` class.
 """
 
-from collections import namedtuple, OrderedDict
+from six import iteritems
+
+from collections import OrderedDict
 
 from math import log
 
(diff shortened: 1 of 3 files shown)


rig/bitfield.py
```diff
@@ -187,41 +203,38 @@ def __call__(self, **field_values):
...
         for identifier, value in field_values.items():
-            field_length = self.fields[identifier].length
+            # All fields specified must exist
+            field = self.fields.get_field(identifier, field_values)
+
+            # All field values must be within range
             if value < 0:
                 raise ValueError("Fields must be positive.")
-            elif field_length is not None and value >= (1 << field_length):
+            elif field.length is not None and value >= (1 << field.length):
```
mundya: In principle this could be:
```python
elif value >= (1 << (field.length or self.length)):
```
For super-high defensiveness...

rig/bitfield.py
```diff
@@ -487,156 +517,374 @@ def __init__(self, length=None, start_at=None, tags=None,
...
-            which will be assigned by bitfields higher up the hierarchy.
+        identifiers : iterable of identifiers
+            The identifiers of the fields to assign
+        field_values : {identifier: value, ...}
+            The values held by various fields (used to access the correct
+            identifiers)
+        assign_positions : bool
+            If False, will only assign fields whose length is unknown (leaving
+            those with undefined positions as they are). If True, will assign
+            positions and lengths as required.
```
mundya: I find the first sentence of this docstring very unclear.

"If False, will only assign lengths to fields whose positions are already known.  Otherwise lengths and positions will be assigned to all fields as necessary." Perhaps?

rig/bitfield.py
mossblaser: Unfortunately it makes behaviour when (unsuported) lengths of 0 are used very odd. At present I think lengths of zero will just work.

Further, it makes the rule of "if you don't have a fixed length, you don't get an exception until you assign lengths" more complicated. I think cases where a bug gets you a value < 1<<32 (typical case) should be rare enough not to warrant extra help.

What do you think?

mundya:
I think I'm good with this... I'd like a run through tomorrow if that's ok?

Otherwise, merge away!
LGTM

PR #98, opened 2015-04-17
Wrappers for commonly used signals

### LINKED ISSUE #79: Easier access to common signals
Certain signals such as sync0 and app_stop particularly deserve easier access. The simplest option is to create trivial wrappers for them.

App-stop may deserve a "with" wrapper, perhaps one which also sets the app_id context...

mundya:
Includes code by @mossblaser.

 - Allows send_signal and count_cores_in_state to optionally accept a string
   naming the signal/state they require.
 - Provides a context manager extension to allow:
```python
with: cn.application(63):
    # app_id is set to 63
    # On leaving this block `app_stop(63)` will be called
```
- Add wait_for_cores_to_reach_state function.

Fixes #79

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -233,6 +233,24 @@ class MachineController(ContextMixin):
         assert len(self.structs) > 0
 
     @ContextMixin.use_contextual_arguments
+    def application(self, app_id=Required):
+        """Update the context to use the given application ID and stop the
+        application when done.
+
+        For example::
+
+            with cn.application(54):
+                # All commands in this block will use app_id=54.
+                # On leaving the block `cn.send_signal("stop", 54)` is
+                # automatically called.
+        """
+        # Get a new context and add a method that will be called before the
+        # context is removed from the stack.
+        context = self(app_id=app_id)
+        context.before_close(lambda: self.send_signal("stop"))
+        return context
+
+    @ContextMixin.use_contextual_arguments
     def get_software_version(self, x=Required, y=Required, processor=0):
         """Get the software version for a given SpiNNaker core.
 
@@ -888,7 +906,7 @@ class MachineController(ContextMixin):
 
         # If not waiting then send the start signal
         if not wait:
-            self.send_signal(consts.AppSignal.start, app_id)
+            self.send_signal("start", app_id)
 
     @ContextMixin.use_contextual_arguments
     def send_signal(self, signal, app_id=Required):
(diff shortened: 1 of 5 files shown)


rig/machine_control/machine_controller.py
```diff
@@ -917,6 +935,30 @@ def send_signal(self, signal, app_id=Required):
...
+        """Transmit a signal to start the application."""
+        self.send_signal(consts.AppSignal.start, app_id)
+
+    @ContextMixin.use_contextual_arguments
+    def app_stop(self, app_id=Required):
+        """Transmit a signal to stop the application."""
+        self.send_signal(consts.AppSignal.stop, app_id)
+
+    @ContextMixin.use_contextual_arguments
+    def sync0(self, app_id=Required):
```
mundya: In principle this could accept a count and wait until the count was hit...?

rig/machine_control/machine_controller.py
```diff
@@ -917,6 +935,30 @@ def send_signal(self, signal, app_id=Required):
         self._send_scp(0, 0, 0, SCPCommands.signal, arg1, arg2, arg3)
 
     @ContextMixin.use_contextual_arguments
+    def app_start(self, app_id=Required):
+        """Transmit a signal to start the application."""
+        self.send_signal(consts.AppSignal.start, app_id)
```
mossblaser: What is app start used for in normal use btw?

rig/machine_control/machine_controller.py
mossblaser: Quite possibly... Raises the issue of having the equivalent "count" utility available too?

rig/machine_control/machine_controller.py
```diff
@@ -233,6 +233,24 @@ def boot(self, width, height, **boot_kwargs):
         assert len(self.structs) > 0
 
     @ContextMixin.use_contextual_arguments
+    def application(self, app_id=Required):
+        """Update the context to use the given application ID and stop the
+        application when done.
+
+        For example::
+
+            cn.application(54):
```
mossblaser: with cn.application(54):

rig/utils/contexts.py
```diff
@@ -167,16 +167,25 @@ def __init__(self, context_arguments, stack=None):
         """
         self.context_arguments = dict(context_arguments)
         self.stack = stack
+        self.before_closing = list()
 
     def update(self, updates):
         """Update the arguments contained within this context."""
         self.context_arguments.update(updates)
 
+    def before_close(self, fn):
```
mossblaser: s/before_close/before_closing/ (?)

(prefix the internal variable with _)

rig/utils/contexts.py
```diff
@@ -167,16 +167,25 @@ def __init__(self, context_arguments, stack=None):
...
+        """Call the given function before this context is exited."""
+        self.before_closing.append(fn)
+
     def __enter__(self):
         # Add this context object to the stack
         self.stack.append(self)
 
     def __exit__(self, exception_type, exception_value, traceback):
+        # Call all the passed functions before closing the context
+        for fn in self.before_closing:
```
mossblaser: This and the stack-poping could be in a try: finally: block to leave things in a consistent state if a before-closing function fails.

rig/machine_control/machine_controller.py
mundya: As in `count_sync0`? I'm not sure I see why that would be particularly useful as separate to this?

mossblaser: Something like that. Though a "wait for count" seems like a frequently useful thing, I can see it as being desirable to also check for failed processes etc. rather than waiting for a timeout or implement some other more customised timeout mechanism. It would also be useful for people using sync0 and 1 as general purpose application barriers.

mundya: Actually, I totally get the use for that - d'oh. There's some really cool stuff we can do here because we can grab the full core state (like write the reason for any RTEs out).  I'll get on this tomorrow.

rig/machine_control/machine_controller.py
mundya: To leave the paused state (see `load_application`)... it's probably not the *most* useful signal to expose like this, but I'm tempted to keep it anyway.

mundya:
Don't merge yet, I'm working on waiting until a count is hit still...

mossblaser:
Could I request the "exit" signal is also broken out?

mundya:
Yes, but I believe you have a better suggestion anyway :wink:

mundya:
This is just awaiting "wait_until_cores_in_state"...

mossblaser:
Just writing the tests :)

mundya:
Ah, patience... I've heard of that

mossblaser:
Feel free to squash :) and merge at will!

mundya:
I'll merge once the test clears... :hourglass_flowing_sand:

PR #100, opened 2015-04-21
`alloc_as_filelike` for vertices

mundya:
Allow allocation of SDRAM for a number of vertices.  `sdram_alloc_as_filelike` now accepts resources, placements and allocations and will perform SDRAM allocation for vertices given this information.  A dictionary mapping vertices to SDRAM file-like objects will be returned.

The older calling convention is still provided.

### DIFF
--- a/rig/machine_control/utils.py
+++ b/rig/machine_control/utils.py
@@ -1,0 +1,87 @@
+import six
+from ..machine import Cores, SDRAM
+
+
+def sdram_alloc_for_vertices(controller, placements, allocations,
+                             core_as_tag=True,
+                             sdram_resource=SDRAM, cores_resource=Cores):
+    """Allocate and return a file-like view of a region of SDRAM for each
+    vertex which uses SDRAM as a resource.
+
+    The tag assigned to each region of assigned SDRAM is the index of the
+    first core that each vertex is assigned.  For example::
+
+        placements = {vertex: (0, 5)}
+        allocations = {vertex: {Cores: slice(3, 6),
+                                SDRAM: slice(204, 304)}}
+        sdram_allocations = sdram_alloc_for_vertices(
+            controller, placements, allocations
+        )
+
+    Will allocate a 100-byte block of SDRAM for the vertex which is
+    allocated cores 3-5 on chip (0, 5).  The region of SDRAM will be tagged
+    `3` (because this is the index of the first core).
+
+    Parameters
+    ----------
+    controller : :py:class:`rig.machine_control.MachineController`
+        Controller to use to allocate the SDRAM.
+    placements : {vertex: (x, y), ...}
+        Mapping of vertices to the chips they have been placed on.  Same as
+        produced by placers.
+    allocations : {vertex: {resource: allocation, ...}, ...}
+        Mapping of vertices to the resources they have been allocated.
+
+        A block of memory of the size specified by the `sdram_resource`
+        (default: :py:class:`~rig.machine.SDRAM`) resource will be allocated
+        for each vertex. Note that location of the supplied allocation is *not*
+        used.
+
+        When `core_as_tag=True`, the tag allocated will be the ID of the first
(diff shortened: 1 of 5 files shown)


rig/machine_control/machine_controller.py
```diff
@@ -640,10 +681,48 @@ def sdram_alloc_as_filelike(self, size, tag=0, x=Required, y=Required,
             If the memory cannot be allocated, or the tag is already taken or
             invalid.
         """
-        # Perform the malloc
-        start_address = self.sdram_alloc(size, tag, x, y, app_id)
-
-        return MemoryIO(self, x, y, start_address, start_address + size)
+        if len(args) == 2:
```
mossblaser: This doesn't seem to allow for optional definition of the tag anymore?

mossblaser:
I like the feature very much, however my first impression is that I feel that this sort of function should probably exist as a separate utility method (or maybe even a standalone utility function?).

By the way, isn't the "resources" dictionary essentially repeating information available in the allocation dictionary or am I missing something?

mundya:
> I like the feature very much, however my first impression is that I feel that this sort of function should probably exist as a separate utility method (or maybe even a standalone utility function?).

I thought this, decided against, and now decide in favour again as:
```python
controller(app_id=xxx):
    some_other_method(controller, ...)
```
Will still use the contexts.

> By the way, isn't the "resources" dictionary essentially repeating information available in the allocation dictionary or am I missing something?

I hadn't thought of extracting the SDRAM usage from the slices.  I shall fix that.

mossblaser:
Perhaps something like:

    def something_something_alloc(placement, allocation, core_as_tag=True,
                                  core_resource=Cores, sdram_resource=SDRAM):
        vertex_files = {}
        for vertex, (x, y) in iteritems(placement):
            slice = allocation[vertex][sdram_resource]
            size = slice.end - slice.start
            if core_as_tag:
                tag = allocation[vertex][core_resource].start
            else:
                tag = 0
            vertex_files[vertex] = machine.sdram_alloc_as_filelike(size, tag,
                                                                   x=x, y=y)

mossblaser:
> I thought this, decided against, and now decide in favour again as:

A method it is :).

Like I say, a separate method just seems somewhat clearer. Also, I think that the "alloc" and "alloc as file" methods should both accept the same type of arguments, whatever way this is done.

rig/machine_control/utils.py
```diff
@@ -0,0 +1,78 @@
+import six
+from ..machine import Cores, SDRAM
+
+
+def sdram_alloc_for_vertices(controller, placements, allocations,
+                             core_as_tag=True,
+                             sdram_resource=SDRAM, cores_resource=Cores):
+    """Allocate a region of SDRAM for each vertex which uses SDRAM as a
+    resource.
```
mossblaser: "and return a MemoryIO object for each vertex" or something :)

rig/machine_control/utils.py
```diff
@@ -0,0 +1,78 @@
...
+
+    Parameters
+    ----------
+    controller : :py:class:`rig.machine_control.MachineController`
+        Controller to use to allocate the SDRAM.
+    placements : {vertex: (x, y), ...}
+        Mapping of vertices to the chips they have been placed on.  Same as
+        produced by placers.
+    allocations : {vertex: {resource: allocation, ...}, ...}
+        Mapping of vertices to the resources they have been allocated.  As
```
mossblaser: Mapping of vertices to the resources they have been allocated.

A block of memory of the size specified by the `sdram_resource` (default:
:py:class:`~rig.machine.SDRAM`) resource will be allocated for each vertex. Note
that location of the supplied allocation is *not* used.

When `core_as_tag=True`, the tag allocated to allocations will be the ID of the
first core used by the vertex, otherwise the tag will be set to 0.

mossblaser:
Above commits add some minor documentation additions (example and types in docstring).

LGTM! Feel free to squash history!

Issue #99, opened 2015-04-20
Data structure handling (a.k.a. 'Regions')


mossblaser:
This issue aims to collect together thoughts on the problem currently solved by
the 'Regions' system and to either better understand it and hopefully improve
it.

# Requirements

**In a nutshell:** to assist in translating between Python data structures on
the host and C data structures spread across the SpiNNaker hardware.

* High-level
  * Translation must generally be bidirectional, though not all applications
    will require translating in both directions! Perhaps "read only", "write
    only" and "read/writable" could be the three possible classes... Perhaps
    not, though... (**Regions do not explicitly represent this information.** Is
    this a bug or a feature?)
  * Data structures on the (single) host may be spread across multiple SpiNNaker
    devices.
  * Data structures will be relocatable (i.e. no absolute pointers) and the API
    should not need to know the locations where generated blocks will be
    written.
* API
  * A common interface should be defined which all data types should obey.
  * Should very cleanly interface with appropriate MachineController methods
    *and* with file-likes (to allow easy dumping/caching) with minimal
    boilerplate code. Minimal magic should be involved! Note that this API
    should ideally allow data to be referred to with minimal unnecessary
    duplication.
  * It should be possible to divide certain data structures (e.g. matrices),
    ideally without duplicating memory.
  * Interface should allow lazy evaluation of structure contents. Note that it
    should be considered acceptable to have a whole SDRAM-worth of data in RAM
    at once.
  * Should provide ready-made implementations of common data structures, a
    complementary (and minimal) C library should be provided for interpreting
    these (to reduce ambiguity), but considered optional and formatting should
    be carefully described in documentation.
  * Should be possible to compose and nest certain data structures
  * Certain compound types should support "slicing" with minimal copying where
    possible.
* (Potential) Included data types
  * Primitives
    * Pointer
    * Char
    * Int/uint/etc.
    * Float/Double
    * Fixed-point
    * String/Bytes
  * Compound
    * (C-style) Struct
    * Array
    * Linked list(?)
    * Key-Value store (with fixed set of keys)
    * Dictionary(?)
  * Numpy
    * Int
    * Float
    * Double
    * Fixed point
    * Matrix
  * Functional
    * Map (e.g. for 'keyspaces')

# Naming

It has been noted that the name 'Region' has been too heavily overloaded and
perhaps doesn't get-across what this module is doing.

Since this functionality principally deals with the problem of data-structure
definition and translation. As a result, I propose the submodule is renamed
[...]

PR #101, opened 2015-04-21
Remove the regions submodule.

mossblaser:
Following #99, it was decided to remove the 'Regions' concept from Rig since it
does not appear to be useful to most Rig projects.


mundya:
LGTM.

PR #102, opened 2015-04-21
Subscriptable MemoryIO

mundya:
Allows MemoryIO to be subscripted to generate new file-likes which reference only a portion of the memory of the original file-like view.

For example:
```python
>>> f = controller.sdram_alloc_as_filelike(12)
>>> f.write(b"Hello, world")
>>> f.read()
b"Hello, world"
>>> g = f[0:5]
>>> g.read()
b"Hello"
>>> f.seek(0)
>>> f.read()
b"Hello, world"
```

### DIFF
--- a/docs/source/control.rst
+++ b/docs/source/control.rst
@@ -175,6 +175,13 @@ allocated region. For example::
     >>> block.read(13)
     b"Hello, world!"
 
+This file-like wrapper is sliceable so that new smaller file-like views of
+memory may be constructed::
+
+    >>> hello = block[0:5]
+    >>> hello.read()
+    b"Hello"
+
 The :py:func:`~rig.machine_control.utils.sdram_alloc_for_vertices` utility
 function is provided to allocate multiple SDRAM blocks simultaneously.  This
 will be especially useful if you're using Rig's :doc:`place and route
@@ -188,7 +195,6 @@ output format. For example::
     
     >>> # The returned dictionary maps from vertex to file-like wrappers
     >>> vertex_memory[vertex].write(b"Hello, world!")
-
 
 Context Managers
 ^^^^^^^^^^^^^^^^
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -1352,7 +1352,21 @@ class SpiNNakerLoadingError(Exception):
 
 
 class MemoryIO(object):
-    """A file-like view into a subspace of the memory-space of a chip."""
+    """A file-like view into a subspace of the memory-space of a chip.
+
+    A `MemoryIO` is sliceable to allow construction of new, more specific,
+    file-like views of memory.
+
+    For example::
+
+        >>> f = MemoryIO(mc, 0, 1, 0x67800000, 0x6780000c)  # doctest: +SKIP
+        >>> f.write(b"Hello, world")                        # doctest: +SKIP
+        >>> f.read()                                        # doctest: +SKIP
+        b"Hello, world"
+        >>> g = f[0:5]                                      # doctest: +SKIP
+        >>> g.read()                                        # doctest: +SKIP
+        b"Hello"
+    """
 
     def __init__(self, machine_controller, x, y, start_address, end_address):
         """Create a file-like view onto a subset of the memory-space of a chip.
@@ -1370,16 +1384,73 @@ class MemoryIO(object):
             Starting address in memory.
         end_address : int
             End address in memory.
+
+        If `start_address` is greater or equal to `end_address` then
+        `end_address` is ignored and `start_address` is used instead.
         """
         # Store parameters
         self._x = x
         self._y = y
         self._machine_controller = machine_controller
+
+        # Store and clip the addresses
         self._start_address = start_address
-        self._end_address = end_address
+        self._end_address = max(start_address, end_address)
 
         # Current offset from start address
         self._offset = 0
+
+    def __getitem__(self, sl):
+        """Get a new file-like view of SDRAM covering the range indicated by
+        the slice.
+
+        For example, if `f` is a `MemoryIO` covering a 100 byte region of SDRAM
+        then::
+
+            >>> g = f[0:10]  # doctest: +SKIP
+
+        Creates a new `MemoryIO` referring to just the first 10 bytes of `f`,
+        the new file-like will be positioned at the start of the given block::
+
+            >>> g.tell()  # doctest: +SKIP
+            0
+
+        Raises
+        ------
+        ValueError
+            If the slice is not contiguous.
+        """
+        if isinstance(sl, slice) and (sl.step is None or sl.step == 1):
+            # Get the start and end addresses
+            if sl.start is None:
+                start_address = self._start_address
+            elif sl.start < 0:
+                start_address = max(self._start_address,
+                                    self._end_address + sl.start)
+            else:
+                start_address = min(self._end_address,
+                                    self._start_address + sl.start)
+
+            if sl.stop is None:
+                end_address = self._end_address
+            elif sl.stop < 0:
+                end_address = max(start_address,
+                                  self._end_address + sl.stop)
+            else:
+                end_address = min(self._end_address,
+                                  self._start_address + sl.stop)
+
+            # Construct the new file-like
+            return type(self)(
+                self._machine_controller, self._x, self._y,
+                start_address, end_address
+            )
+        else:
+            raise ValueError("Can only make contiguous slices of MemoryIO")
+
+    def __len__(self):
+        """Return the number of bytes in the file-like view of SDRAM."""
+        return self._end_address - self._start_address
 
     def read(self, n_bytes=-1):
         """Read a number of bytes from the memory.
@@ -1406,8 +1477,8 @@ class MemoryIO(object):
         if self.address + n_bytes > self._end_address:
             n_bytes = min(n_bytes, self._end_address - self.address)
 
-            if n_bytes <= 0:
-                return b''
+        if n_bytes <= 0:
+            return b''
 
         # Perform the read and increment the offset
         data = self._machine_controller.read(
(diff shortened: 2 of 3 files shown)


mundya:
Note: fixes #68.

docs/source/control.rst
```diff
@@ -175,6 +175,13 @@ allocated region. For example::
     >>> block.read(13)
     b"Hello, world!"
 
+This file-like wrapper is subscriptable so that new smaller file-like views of
+memory may be constructed::
```
mossblaser: This file-like wrapper is **sliceable** so that new smaller file-like views of **subsets** of memory may be constructed::

rig/machine_control/machine_controller.py
```diff
@@ -1352,7 +1352,21 @@ def __str__(self):
 
 
 class MemoryIO(object):
-    """A file-like view into a subspace of the memory-space of a chip."""
+    """A file-like view into a subspace of the memory-space of a chip.
+
+    A `MemoryIO` is subscriptable to allow construction of new, more specific,
```
mossblaser: s/subscriptable/sliceable/

rig/machine_control/machine_controller.py
```diff
@@ -1352,7 +1352,21 @@ def __str__(self):
...
+    file-like views of memory.
+
+    For example::
+
+        >>> f = MemoryIO(mc, 0, 1, 0x67800000, 0x6780000c)  # doctest: +SKIP
```
mossblaser: An informal convention I've seen around is to use a number of cheverons != 3 which makes doctest ignore a block. This seems somewhat less ugly than the spamming of doctest comments. Sigh, how *nearly* awesome doctest is...

rig/machine_control/machine_controller.py
```diff
@@ -1381,6 +1395,48 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+        ------
+        ValueError
+            If the slice is not contiguous.
+        """
+        if isinstance(sl, slice) and (sl.step is None or sl.step == 1):
+            # Get the start and end addresses
+            if sl.start is None:
+                start_address = self._start_address
+            elif sl.start < 0:
+                start_address = max(0, self._end_address + sl.start)
```
mossblaser: The above may allow a start_address which is < the original start address (i.e. to grow the range of memory the MemoryIO can access). This would be a bug in my eyes.

rig/machine_control/machine_controller.py
```diff
@@ -1381,6 +1395,48 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+            else:
+                start_address = min(self._end_address,
```
mossblaser: Since the end address may be changed too, this may result in the start_address being after the new end_address.

docs/source/control.rst
mundya: subscriptable is an appropriate term here... but yeah, slicing is nicer

rig/machine_control/machine_controller.py
mundya: Damn, `start_address = max(self._start_address, ...)`

rig/machine_control/machine_controller.py
mundya: I've seen that also, I'm not sure what Sphinx makes of it... let's see...

rig/machine_control/machine_controller.py
mundya: > this would result in silently creating a "0 length" slice which is the same behaviour as slicing in Python generally.

Probably best not make that silent (i.e., explicitly clip the addresses anyway).

rig/machine_control/machine_controller.py
```diff
@@ -1381,6 +1395,48 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+                                    self._start_address + sl.start)
+
+            if sl.stop is None:
+                end_address = self._end_address
+            elif sl.stop < 0:
+                end_address = max(0, self._end_address + sl.stop)
```
mossblaser: Is it necessary to clamp at 0 here?

docs/source/control.rst
mundya: Disagree on subsets, sorry.

rig/machine_control/machine_controller.py
mundya: Sphinx doesn't format this with only two chevrons... comments or chevrons: your thoughts please...

mossblaser:
Random thought: I think adding `__len__` would be welcome while you're at this!

As noted in comments, I'm not convinced about the logic being used to clamp the address ranges of the new MemoryIO. Plus I think there should be some tests for MemoryIOs of length 0 to make sure behaviour is safe & defined.

Aside from that :) I like! :D

rig/machine_control/machine_controller.py
mossblaser: Gah! stupid doctest! Comments I think... Aligned off to the right though, perhaps, to minimise visual noise?

rig/machine_control/machine_controller.py
mundya: > Either way, the "min" here is doing something either pointless or wrong.
How so?  The new start address must be in the range old_start < start <= old_end.

mossblaser: It should be silent:

    >>> a = list(range(10))
    >>> a[5:3]
    []

But yes, the bounds *must* be clipped so that the start and end end up equal otherwise truncation logic in the MemoryIO block looks like it would fail.

mossblaser: > How so? The new start address must be in the range old_start < start <= old_end.

Surely old_start <= start < old_end ?

mundya: > Surely old_start <= start < old_end ?

Can we agree on old_start <= start <= old_end, where start == old_end -> empty?

mossblaser: > Can we agree on old_start <= start <= old_end, where start == old_end -> empty?

old_start <= start <= min(old_end, end), where start == end -> empty

rig/machine_control/machine_controller.py
mundya: Done.

mossblaser:
Logic looks good now, though I'd be comfortable if there were a handful of tests covering cases where the resulting MemoryIO had zero length. (That and some tests for zero-length MemoryIOs).

mundya:
> though I'd be comfortable if there were a handful of tests covering cases where the resulting MemoryIO had zero length. (That and some tests for zero-length MemoryIOs).

On the way (along with `__len__`)

mossblaser:
> On the way (along with `__len__`)

Schweet. Patience is a virtue I seem to have lost ;).

mundya:
With your permission I'll squash.

rig/machine_control/machine_controller.py
```diff
@@ -1375,12 +1389,62 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+        # Store and clip the addresses
         self._start_address = start_address
-        self._end_address = end_address
+        self._end_address = max(start_address, end_address)
 
         # Current offset from start address
         self._offset = 0
 
+    def __getitem__(self, sl):
+        """Get a new file-like view of SDRAM covering the range indicated by
```
mossblaser: It would be worth explicitly documenting what the offset is after splitting. I'm not sure which is better out of "always 0" or "the old offset shifted and clamped to the new range". I can't see the latter being too useful, except possibly if you are truncating something... but that seems a little dubious. Either way, it should be documented.

mundya: This is fixed slightly further down in this docstring.

mundya:
@mossblaser has, I believe, given the go-ahead for a merge when this test passes.

mossblaser:
Correct :) All LGTM :).

mundya:
Erm... this appears to have gone missing again.

mundya:
Should be back... Not sure what's going on here,

mossblaser:
I don't think its me this time... I hope...

mossblaser:
In case we get into this in the future... the following (totally against github usage policy) shell script will check that every pull request is still merged into the current branch.

```bash
#!/bin/bash

# Run in the rig git repository with arguments as the range of issue/pr numbers
# to scan. For each PR, it is checked to see that the branch is merged into the
# currently checked out branch.
#
# e.g. checks all PRs 1-123
#    ./gitcheck.sh 123
# e.g. checks all PRs 100-123
#    ./gitcheck.sh 100 123

# Branch into which we care about merges
target_branch="$(git branch | grep "*" | cut -f2 -d" ")"

branch_shas="$(git log --pretty=oneline | cut -f1 -d" ")"

function checkmerged() {
	num="$1"
	while ! page="$(curl -m 5 "https://github.com/project-rig/rig/pull/$num" -s)"; do
		sleep 1
	done
	
	# Ignore issues
	if echo "$page" | grep -qE "You are being .*redirected"; then
		echo "$num: is not a pull request"
		return 0;
	fi
	
	# Grab the commit SHA
	hash="$(echo "$page" | sed -nre 's:.*merged commit <a href="/project-rig/rig/commit/([0-9a-fA-F]+)">.*</a> into <span.*>'"$target_branch"'<.*:\1:p')"
	
	# Ignore if not merged
	if [ -z "$hash" ]; then
		echo "$num: not merged yet"
		return 0;
	fi
	
	if echo "$branch_shas" | grep -q "$hash"; then
		echo "$num: still merged"
		return 0
	else
		echo "$num: WARNING: nolonger merged into branch $target_branch!"
		return 1
	fi
}

for num in `seq $@`; do
	checkmerged $num
done
```

mundya:
You could add this to the `utils` directory in the root of the repository? (Where `rig_path` lives).

mossblaser:
I'd rather not make this an "official" tool. This was a silly situation
which shouldn't have occurred and the above tool should really be for a
"dire emergency" sort of situation only.

mundya:
Fair enough, thank you.

PR #103, opened 2015-04-21
Remove Numpy as a requirement

mundya:
I'm not sure how we want to test this - we could create a new Tox environment that doesn't install Numpy and check that the method isn't available...?

### DIFF
--- a/rig/type_casts.py
+++ b/rig/type_casts.py
@@ -1,6 +1,9 @@
 """Fixed point conversion utilities.
 """
-import numpy as np
+try:
+    import numpy as np
+except ImportError:
+    np = None  # No Numpy support
 
 
 def float_to_fix(signed, n_bits, n_frac):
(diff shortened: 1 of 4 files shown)


rig/type_casts.py
```diff
@@ -264,3 +178,97 @@ def validate_fp_params(signed, n_bits, n_frac):
     max_v = ((1 << n_int) - 1) / float(1 << n_frac)
 
     return min_v, max_v
+
+
+if np is not None:
```
mossblaser: # pragma: no branch

rig/type_casts.py
```diff
@@ -1,6 +1,9 @@
 """Fixed point conversion utilities.
 """
-import numpy as np
+try:
+    import numpy as np
+except ImportError:
+    np = None  # No Numpy support
```
mossblaser: # pragma: no cover

setup.py
```diff
@@ -12,8 +12,9 @@
     license="GPLv2",
 
     # Requirements
-    install_requires=["numpy>1.6", "six", "enum34", "sentinel"],
+    install_requires=["six", "enum34", "sentinel"],
     tests_require=["pytest>=2.6", "pytest-cov", "mock", "toposort"],
```
mossblaser: tests need "numpy>1.6"

mossblaser:
An alternative for could be to monkeypatch numpy out in the tests file and re-import the module and verify the functions are missing in a test. That way you don't need a whole test run to check the behaviour.

Incidentally, might it not be better from the principle of least surprise point of view to replace numpy-dependencies with stubs which produce a suitably informative exception? (Could be done with a decorator too, and everyone loves decorators!)

mundya:
Decorator sounds good, I'll get on that.

mossblaser:
Mutually agreed that this idea will be shelved for now due to pains in testing etc.

PR #105, opened 2015-04-22
Fix function signatures when using contexts

### LINKED ISSUE #75: Autodocumentation for context supplied arguments
@mossblaser said:

> I'd also like it if the use named contextual arguments could update the function signature too for auto-documentation purposes.

Modifying the function signature seems like a Very Bad<sup>TM</sup> / difficult thing to do.  My proposal would be to modify the docstring by automagically adding in the context kwargs.

Any thoughts?

mossblaser:
Adds a new decorator which adds a rendering of a function's type to the top of a
wrapper function's docstring. This line is used by Sphinx autodoc to generate
the function signature in the documentation. This can also be used on Python 2
to work-around the fact that functools.wrapper doesn't preserve the underlying
function's prototype.

Fixes #75


mossblaser:
[Live demo in RTD](http://rig.readthedocs.org/en/fix-doc-function-signatures/control_api.html#rig.machine_control.MachineController.send_scp)

[For comparison, here is master -- note *all* context-aware function signatures are broken](http://rig.readthedocs.org/en/master/control_api.html#rig.machine_control.MachineController.send_scp)

mossblaser:
Coverage... Whoops!

mundya:
Docs are significantly nicer :)

mundya:
LGTM, I think we should go through at some point and add all the context aware arguments as something like:

```
Optional Arguments
------------------
x : int (context-supplied)
```

But that's for another day.

mossblaser:
Code coverage should now be fixed: I was unnecessarily checking for args being None and all the example functions needed pragma-no-covering.

> I think we should go through at some point and add all the context aware arguments

Possibly, though since they're very common (and in many cases, obvious), the only critical thing is that their existence, required-ness and position are documented. More pressingly, I'm very lazy.

PR #107, opened 2015-04-23
Accept string signal/state names.

mossblaser:
Removes wrappers around send_signal in favour of allowing send_signal and
count_cores_in_state to optionally accept a string naming the signal/state they
require.

(Note this PR merges into #98, not master :))


mundya:
LGTM.

PR #108, opened 2015-04-23
Add wait_for_cores_to_reach_state function.

mossblaser:
Also flake8-ify.

Also merges into #98 which in turn should be mergeable if this is OK!

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -1045,7 +1045,7 @@ class MachineController(ContextMixin):
             if cur_count >= count:
                 break
 
-            # Stop if timeout ellapsed
+            # Stop if timeout elapsed
             if timeout is not None and time.time() > timeout_time:
                 break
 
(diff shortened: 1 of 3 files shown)


mossblaser:
Should probably add some tutorial docs... Getting on that. Don't merge just yet!

rig/machine_control/machine_controller.py
```diff
@@ -995,6 +996,65 @@ def count_cores_in_state(self, state, app_id=Required):
...
+        """
+        if timeout is not None:
+            timeout_time = time.time() + timeout
+
+        while True:
+            cur_count = self.count_cores_in_state(state, app_id)
+            if cur_count >= count:
+                break
+
+            # Stop if timeout ellapsed
```
mundya: sp. elapsed

mundya:
I'd definitely like something that included a: "is anything RTEd" check... but I guess that's trivial to add later.

mossblaser:
Yeah... this is true, though as you say, I think I'd rather add this later since I'm not sure exactly how it should look.

Assuming these docs don't break the tests, everything should be good to go, I think!

PR #109, opened 2015-04-23
Allow context to set all arguments

mossblaser:
This commit makes the following contributions:
* All arguments in a context-using function can be set, not just ones with
  default values and those given a the Required sentinel as default.
* The 'Required' sentinel can now be omitted from code in a large number of
  cases, improving readability. This has been done where possible in the
  MachineController and BMPController codebase.
* The use_contextual_arguments and use_named_contextual_arguments decorators
  have been merged together into use_contextual_arguments which now takes
  kwargs in the way use_named_contextual_arguments used to. If no arguments are
  given, the behaviour is the same as 'use_contextual_arguments' used to be.

### DIFF
--- a/rig/machine_control/bmp_controller.py
+++ b/rig/machine_control/bmp_controller.py
@@ -116,7 +116,7 @@ class BMPController(ContextMixin):
         """Create a new context for use with `with`."""
         return self.get_new_context(**context_args)
 
-    @ContextMixin.use_named_contextual_arguments(
+    @ContextMixin.use_contextual_arguments(
         cabinet=Required, frame=Required, board=Required)
     def send_scp(self, *args, **kwargs):
         """Transmit an SCP Packet to a specific board.
@@ -168,9 +168,8 @@ class BMPController(ContextMixin):
 
         return connection.send_scp(length, 0, 0, board, *args, **kwargs)
 
-    @ContextMixin.use_contextual_arguments
-    def get_software_version(self, cabinet=Required, frame=Required,
-                             board=Required):
+    @ContextMixin.use_contextual_arguments()
+    def get_software_version(self, cabinet, frame, board):
         """Get the software version for a given BMP.
 
         Returns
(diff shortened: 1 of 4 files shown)


rig/machine_control/bmp_controller.py
```diff
@@ -168,9 +168,8 @@ def _send_scp(self, cabinet, frame, board, *args, **kwargs):
 
         return connection.send_scp(length, 0, 0, board, *args, **kwargs)
 
-    @ContextMixin.use_contextual_arguments
-    def get_software_version(self, cabinet=Required, frame=Required,
-                             board=Required):
+    @ContextMixin.use_contextual_arguments()
```
mundya: I dislike the mandatory brackets... but, if I remember correctly, this is a hard problem to fix.

mossblaser: Yeah :/ this is a shame...

One way to fix it is to detect if you've received a single function as the only argument, and if so, invoke the decorator... It suddenly becomes a bit bizarre though since it breaks consistency and makes everything rather more magic than it needs to be.

mundya: It doesn't _look_ magic from the coder point of view though, I don't think.  Presumably:

```python
def use_contextual_arguments(f=None, **kwarg_only_args_defaults):
    def decorator(f):
        # ...
    if f is not None:
        return decorator(f)
    else:
        return decorator
```

Would do it?
```

Comment (mundya) [inline rig/machine_control/bmp_controller.py]:
```python
def use_contextual_arguments(*args, **kwarg_only_args_defaults):
    def decorator(f):
        # ...
    if len(args) == 1:
        return decorator(args[0])
    else:
        return decorator
```

mossblaser: With the minor issue that suddenly now you can confuse yourself (silently) with:

	@use_contextual_arguments(my_mandatory_kw_only_argument)
	def f(**kwargs):
		pass

But anyway, when a problem with decorators arrises, just add decorators!

def maybe_args_decorator(d):
	@functools.wraps(d)
	def f_(*args, **kwargs):
		if len(args) == 1 and hasattr(args[0], "__call__") and not kwargs:
			return d()(args[0])
		else:
			return d(*args, **kwargs)

(Untested)

PR #112, opened 2015-04-24
Run flake8 in pytest

mundya:
- Adds `pytest-flake8` as a dependency
 - Includes `--flake8` in pytest invocation
 - Ignores pep8 infractions in docs
 - Ignores `F401` in `__init__.py` files
 - Removes separate flake8 run from tox

Modifies `conftest.py` to avoid failing on the inserted Flake8 test item.


mundya:
For reference F401 is importing something and then not using it, something `__init__.py`s can legitimately do.

mossblaser:
As noted out-of-band, damn: those are some slow travis run times :(. @mundya do you fancy trying the tox method?

PR #114, opened 2015-04-25
SCP Performance Improvements

mundya:
- Pushes `read` and `write` functionality down to the level of the `SCPController` (8116dcd)
 - Removes the value and type checking from SCP and SDP packets (fa05d4e)
 - Slightly neatens SDP internals (1d14607)

Reading and writing 10MB data to chip (0, 0) on a SpiNN3 board I have at home gives:

Commit | Write speed / Mbs<sup>-1</sup> | Read speed / Mbs<sup>-1</sup>
------ | ------------------------------ | -----------------------------
Master | 5.41 | 5.01
After 8116dcd | 5.80 | 5.30
After 8116dcd and fa05d4e | 8.70 | 7.80
After all | 8.90 | 8.00

~~Obviously the results are a little rough, but being faster than @mossblaser's profiling of ybug is pleasing :)~~ Comparable to ybug in the same tests, which is probably a good place to be.

@mossblaser - If you wanted to be nerd-sniped then introducing trollius back into the equation might be interesting.

 - [x] Move the packing bytes into the SDP header struct packing
 - [x] Investigate improving the efficiency of packing SDP packets (particularly fewer string add operations and more memoryview/bytearray writes).

### DIFF
--- a/rig/machine_control/consts.py
+++ b/rig/machine_control/consts.py
@@ -304,3 +304,12 @@ class BMPInfoType(enum.IntEnum):
     can_status = 2  # Status of all CAN devices on the bus
     adc = 3  # ADC (e.g. voltage + temperature)
     ip_addr = 4  # IP Address
+
+
+# Dictionary of (address % 4, n_bytes % 4) to data type
+address_length_dtype = {
+    (i, j): (DataType.word if (i == j == 0) else
+             (DataType.short if (i % 2 == j % 2 == 0) else
+              DataType.byte))
+    for i in range(4) for j in range(4)
+}
--- a/rig/machine_control/packets.py
+++ b/rig/machine_control/packets.py
@@ -1,90 +1,19 @@
 """Representations of SDP and SCP Packets."""
-import six
 import struct
-from weakref import WeakKeyDictionary
 
 # SDP header flags
 FLAG_REPLY = 0x87
 FLAG_NO_REPLY = 0x07
 
 
-class RangedIntAttribute(object):
-    """Descriptor that ensures values fit within a range of values."""
-    def __init__(self, minimum, maximum, min_inclusive=True,
-                 max_inclusive=False, allow_none=False, accept=list()):
-        """Create a new ranged descriptor with specified minimum and maximum
-        values.
-        """
-        if not min_inclusive:
-            minimum += 1  # Convert from [x, y) to (x, y)
-        if max_inclusive:
-            maximum += 1  # Convert from [x, y) to [x, y]
-
-        if not minimum <= maximum:
-            raise ValueError('min should be smaller than max')
-
-        self.accept = accept[:]
-        self.default = minimum
-        self.min = minimum
-        self.max = maximum
-        self.data = WeakKeyDictionary()
-        self.allow_none = allow_none
-
-    def __get__(self, instance, owner):
-        # Get the value, or the default
-        return self.data.get(instance, self.default)
-
-    def __set__(self, instance, value):
-        if value is not None:
-            # Check that min <= value < max
-            if not isinstance(value, six.integer_types):
-                raise TypeError("Value should be an integer: {!s}"
-                                .format(value))
-            if not self.min <= value < self.max and value not in self.accept:
-                raise ValueError("Value outside range {}: [{} to {})"
-                                 .format(value, self.min, self.max))
-            self.data[instance] = value
-        else:
-            if self.allow_none:
-                self.data[instance] = None
-            else:
-                raise ValueError("Value may not be None")
-
-
-class ByteStringAttribute(object):
-    """Descriptor that ensures values are bytestrings of the correct length."""
-    def __init__(self, default=b'', max_length=None):
-        self.max_length = max_length
-        self.default = default
-        self.data = WeakKeyDictionary()
-
-    def __get__(self, instance, owner):
-        return self.data.get(instance, self.default)
-
-    def __set__(self, instance, value):
-        if self.max_length is not None and len(value) > self.max_length:
-            raise ValueError(
-                "Byte string is too long: should be less than {} bytes"
-                .format(self.max_length)
-            )
-        self.data[instance] = value
-
-
 class SDPPacket(object):
     """An SDP Packet"""
-    tag = RangedIntAttribute(0, 256)
-    dest_port = RangedIntAttribute(0, 8)
-    dest_cpu = RangedIntAttribute(0, 18, accept=[31])  # For IPtags
-    src_port = RangedIntAttribute(0, 8)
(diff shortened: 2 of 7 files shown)


mossblaser:
Sounds very good from the numbers!

Have you got a cheap switch you could attach your board via for a more consistent comparison? Out of sheer curiosity, could you try the same with 1,1 too?

As for windowing, if I do do it, it'll be with raw select... trollius is slow and doesn't do very much for you close to the socket. But yes, I could be sniped! :P.

mundya:
> Have you got a cheap switch you could attach your board via for a more consistent comparison? Out of sheer curiosity, could you try the same with 1,1 too?

I don't have a switch, but I can get some numbers for (1, 1)

> As for windowing, if I do do it, it'll be with raw select... trollius is slow and doesn't do very much for you close to the socket. But yes, I could be sniped! :P.

Probably more worth the time interfacing with rig-scp.

mundya:
I hadn't realised quite how slow SCP was across the board.

Reading and writing 10MB data to chip (1, 1)

Commit | Write speed / Mbs<sup>-1</sup> | Read speed / Mbs<sup>-1</sup>
------ | ------------------------------ | -----------------------------
Master | 3.1 | 3.0
Tip of this branch | 4.1 | 4.0

That's some serious overhead.

mundya:
For bonus points this knocks around a minute off the Travis times for Python 3.

mossblaser:
Bugger me, that is serious re:travis!

Could you compare performance with ybug on your setup too for curiosity's
sake?

mundya:
Dubious figures:
ybug write to (0, 0) comes in at ~8.4 Mb/s
ybug read from (0, 0) comes in at ~8.0 Mb/s

Which I'm going to interpret (given the dubious testing) as being in the same ballpark.

---

```bash
$ time echo "sload ..." | ybug ...
$ time echo "sdump ..." | ybug ...
```

mossblaser:
Super! How CPU bound does it look now? May still be worth doing the parallel stuff on python... No point moving to c prematurely! Too much FAFF.

mundya:
> May still be worth doing the parallel stuff on python...

I was tempted to play with writing some code for this this afternoon... because I've nerdsniped myself.

> How CPU bound does it look now?
:cold_sweat: I can't think of a way of measuring that nicely...  suggestions?

mossblaser:
Have fun! Two implementations is enough for me to have written!

Use cprofile to see if most time is spent in the recv function, if so
you're IO bound.

mundya:
> Use cprofile to see if most time is spent in the recv function, if so you're IO bound.

Ok, I realised after panicking that this would be a sane approach.  The difference between communicating with (0, 0) and (1, 1) seems to indicated IO boundedness in the majority of cases anyway...

mundya:
> Have fun!

I shall, but in another PR.  I leave reviewing this one for when you're not supposed to be having a break!

rig/machine_control/scp_connection.py
```diff
@@ -138,6 +138,90 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
                 self.n_tries)
         )
 
+    def read(self, buffer_size, x, y, p, address, length_bytes):
+        """Read a bytestring from an address in memory.
```
mossblaser: A brief note to explain *why* would be welcome perhaps?

rig/machine_control/scp_connection.py
```diff
@@ -138,6 +138,90 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+        return bytes(data)
+
+    def write(self, buffer_size, x, y, p, address, data):
+        """Write a bytestring to an address in memory.
+
+        Parameters
+        ----------
+        buffer_size : int
+            Number of bytes held in an SCP buffer by SARK, determines how many
+            bytes will be expected in a socket.
```
mossblaser: determines how many bytes will be sent per packet?

mundya:
- [ ] Move the packing bytes into the SDP header struct packing
 - [ ] Investigate improving the efficiency of packing SDP packets (particularly fewer string add operations and more memoryview/bytearray writes).

rig/machine_control/scp_connection.py
mundya: Done.  I _hate_ RST.

mundya:
I'll have a final fiddle tomorrow before cleaning up the history a bit.

mundya:
I think I'm at the stage where the phase of the moon is affecting my timing results... so I leave this to @mossblaser to review.

mundya:
For reference, on a locally connected spin3 board I'm reliably hitting 8-10Mib/s on 10MiB transfers to chip (0, 0).  (This is faster than ybug).

rig/machine_control/consts.py
```diff
@@ -304,3 +304,12 @@ class BMPInfoType(enum.IntEnum):
     can_status = 2  # Status of all CAN devices on the bus
     adc = 3  # ADC (e.g. voltage + temperature)
     ip_addr = 4  # IP Address
+
+
+# Dictionary of (address % 4, n_bytes % 4) to data type
+address_length_dtype = {
```
mossblaser: I'm still not convinced I understand why this has to be a dictionary rather than just including the logic in the code (or a function). Is it measurably faster? It implies rather more magic than it implements, that's all... (e.g. as a function: https://github.com/project-rig/rig-scp/blob/master/lib/rs__scp.c#L36-L45)

rig/machine_control/packets.py
```diff
@@ -220,43 +103,71 @@ def from_bytestring(cls, scp_packet, n_args=3):
...
+        if self.arg2 is not None:
+            scp_header += struct.pack('<I', self.arg2)
+        if self.arg3 is not None:
+            scp_header += struct.pack('<I', self.arg3)
 
         # Return the SCP header and the rest of the data
         return scp_header + self.data
+
+
+def _unpack_sdp_into_packet(packet, bytestring):
```
mossblaser: This appears as if it would be sufficiently generic that it should be public(?)

mossblaser: Oh wait, I'm an idiot, it is already accessible. Carry on! :P

mossblaser:
Looks like great work to me.

Rigged

rig/machine_control/consts.py
mundya: I haven't benchmarked this. I'll give that a go later. My only thoughts are that Python's hashing is known awesome.

mundya: As far as I can tell the dictionary is just under twice the speed of calling a method to compute this.

PR #115, opened 2015-04-25
WIP Even faster SCP.

mundya:
Uses sliding window protocol to increase speed for bursts of SCP
messages.

Talking to a locally connected SpiNN3 I'm hitting 19-20 Mb/s writes and
14-15 Mb/s reads (chip 0, 0).  I don't know what SCAMP's support for
multiple outstanding requests on other chips is but the performance is
flaky for me (still, somewhere around 14Mb/s on writes).

 - [x] Neaten documentation
 - [x] Rework `send_scp` to send a burst of one message (mostly to
       reduce duplication of code)
 - [ ] Check on SCAMP status
 - [x] Fill in a few gaps in the tests.
 - [x] Error handling in `send_scp_burst` 
 - [ ] Query window size from machine (thank you @mossblaser)


mossblaser:
Don't forget that you need to query the machine for how many "transient ip tags" it has since this is the hard limit on simultaneous requests.

Great news on those figures! That's dedicatedly a worthwhile improvement! (also rather relieved that C can still beat it ;) ).

My experience has been non 0,0 gets rather unreliable (crashy). 140 hopes to fix this...

mossblaser:
Are your numbers here for window size 5? Should be able to do 8 on current scamp, this did improve things quite a bit for the C version.

rig/machine_control/machine_controller.py
```diff
@@ -131,6 +132,17 @@ def scp_data_length(self):
...
 
+    @property
+    def scp_window_size(self):
+        """The maximum number of packets that can be sent to a SpiNNaker board
+        without receiving any acknowledgement packets.
+        """
+        # If not known, return the default
+        # TODO: Query the machine
+        if self._window_size is None:
+            return 5
```
mossblaser: If in doubt, should be 1, but I'm sure you know that :)

rig/machine_control/scp_connection.py
```diff
@@ -138,7 +160,124 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+        queued_packets = True
+        outstanding_packets = {}
+
+        # While there are packets in the queue or packets for which we are
+        # still awaiting returns then continue to loop.
+        while queued_packets or outstanding_packets:
+            # If there are fewer outstanding packets than the window can take
+            # then transmit a packet and add it to the list of outstanding
+            # packets.
+            if len(outstanding_packets) < window_size:
```
mundya: Oops...

```python
if len(outstanding_packets) < window_size and queued_packets:
```

I suspect an exception being raised for the last few times round this loop isn't good.

mossblaser:
Ah before I forget, you also need to set the transient ip tag timeout to something like a ms or so (rather than the default of seconds) so that a single inter chip packet getting lost doesn't bring everything creating down. This can be done via some IP tag related command.

mundya:
> Great news on those figures! That's dedicatedly a worthwhile improvement! (also rather relieved that C can still beat it ;) ).

I'm sure I can eek out a _bit_ more, but probably not enough to rival the C :wink:

rig/machine_control/machine_controller.py
mundya: Yeah... need to read my SCP manual for the correct way to query this.

/hah a manual!

mundya:
> Are your numbers here for window size 5? Should be able to do 8 on current scamp, this did improve things quite a bit for the C version.

My numbers for (0, 0) are with window size 8.  Because it fell over horribly with anything other than (0, 0) I tried different window sizes.

For additional fun `send_scp_burst` could be used with flood-fill... not that 32KiB cost us all that much time.

mossblaser:
> For additional fun send_scp_burst could be used with flood-fill... not
that 32KiB cost us all that much time.

This is true. In principle there are a good few functions that could be
improved by windowing. Sadly, the "right way" to do this is probably to use
trollius since it quickly becomes hard to manage the interdependencies of
the higher-up levels of abstraction. Since this is probably not worth the
overhead, I'd just leave it at R/W at the moment since that is the only
place where a doubling in speed results in doubling of user happiness ;).

mundya:
As of some more minor tweaking (and removing some debug lines - d'oh) I'm getting around 20Mib/s for both read and write to chip (0, 0) on a locally connected spin3 board.

mundya:
Ug... that Travis fail is a timeout :(

mundya:
Rebasing to get the latest improvements from #114 wasn't happy, so I've created a new branch and PR.

PR #117, opened 2015-04-27
Windowed SCP for reads and writes

mundya:
Uses sliding window protocol to increase speed for bursts of SCP messages.

As of 14th May this results in reads and writes to chip (0, 0) of around 30+Mbit/s when using a window size of 8.  When using a window size of 1 this drops to (a still respectable) 12ish Mbit/s on my test machine.

 - [x] Neaten documentation
 - [x] Rework `send_scp` to send a burst of one message (mostly to
       reduce duplication of code)
 - [x] Fill in a few gaps in the tests.
 - [x] Set the default window size to 1
 - [ ] ~~Correctly query the window size from SC&MP~~ For a later PR
 - [ ] ~~Set the transient IPTag timeout correctly~~ For a later PR


mundya:
Looks like this will be plagued by timeouts...

rig/machine_control/machine_controller.py
```diff
@@ -132,6 +132,17 @@ def scp_data_length(self):
...
 
+    @property
+    def scp_window_size(self):
+        """The maximum number of packets that can be sent to a SpiNNaker board
+        without receiving any acknowledgement packets.
+        """
+        # If not known, return the default
+        # TODO: Query the machine
+        if self._window_size is None:
+            return 5
```
mundya: Default should be 1, should actually query the machine to determine what this value is.

rig/machine_control/scp_connection.py
```diff
@@ -219,19 +381,31 @@ def write(self, buffer_size, x, y, p, address, data):
...
-            dtype = consts.address_length_dtype[(address % 4, block_size % 4)]
+        # Run the event loop and then return the retrieved data
+        self.send_scp_burst(buffer_size, window_size, packets(address, data))
 
-            self.send_scp(buffer_size, x, y, p, consts.SCPCommands.write,
-                          address, block_size, dtype, block)
 
-            address += block_size
-            pos += block_size
+def seqs(mask=0xf):
```
mundya: Default mask can, I think, be safely `0xffff` (and should be).

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +124,169 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+                # There wasn't a returned packet, we may spend quite some time
+                # here.
+                ack = None
+
+            # Process the received packet (if there is one)
+            if ack is not None:
+                # Extract the sequence number from the bytestring, iff possible
+                seq_bytes = ack[2 + consts.SDP_HEADER_LENGTH:
+                                2 + consts.SDP_HEADER_LENGTH + 4]
+                rc, seq = struct.unpack("<2H", seq_bytes)
```
mundya: `unpack_from` would make sense here.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +124,169 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
-                return scp
-
-        # The packet we transmitted wasn't acknowledged.
-        raise TimeoutError(
-            "Exceeded {} tries when trying to transmit packet.".format(
-                self.n_tries)
-        )
-
-    def read(self, buffer_size, x, y, p, address, length_bytes):
+                    outstanding.callback(ack_packet)
```
mundya: We should switch to passing back the bytestring object, then it can be unpacked by the caller if necessary.

mundya:
With tonight's changes this is pushing nearer to 27+MiB/s on 10MiB reads and writes to a locally connected Spin3 board.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+                if (current_time - outstanding.time_sent >
+                        self.default_timeout + outstanding.extra_timeout):
+                    # This packet has timed out, if we have sent it more than
+                    # the given number of times then raise a timeout error for
+                    # it.
+                    if outstanding.n_tries >= self.n_tries:
+                        raise TimeoutError(self.n_tries)
+
+                    # Otherwise we retransmit it
+                    self.sock.send(b"\x00\x00" + outstanding.packet)
```
mundya: This is not good...

tests/machine_control/test_scp_connection.py
```diff
@@ -32,96 +34,273 @@ def mock_conn():
...
+        with pytest.raises(scp_connection.TimeoutError):
+            mock_conn.send_scp_burst(512, 8, packets())
+        fail_time = time.time() - start
+
+        # Failing to transmit should take some time
+        assert fail_time >= mock_conn.n_tries * mock_conn.default_timeout
+
+        # We shouldn't have transmitted the packet more than the number of
+        # times we specified.
+        assert mock_conn.sock.send.call_count == mock_conn.n_tries
```
mundya: Should test that the same thing is transmitted each time!

mundya:
Cancelling running tests is a bad thing that I shouldn't do...

mossblaser:
Is it a bad thing?

mundya:
With the window size set to 1 this branch gets reads and writes at around 11-12Mbit/s against my local board (0, 0) and 5Mbit/s against (1, 1).  (Master is at around 8-9Mbit/s and 4Mbit/s.)

@mossblaser do we want to consider merging this now with the window size set to 1?  We can investigate setting the TTO and reading the window size in an additional PR.

mossblaser:
Via a switch on my machine, window size = 1 I get:

```
Finished write after 11422 ms. Throughput = 7.00 Mbit/s
Finished read after 11520 ms. Throughput = 6.94 Mbit/s
```

Just upping the window now...

mossblaser:
..and with window size 8:

```
Finished write after 2441 ms. Throughput = 32.78 Mbit/s
Finished read after 2404 ms. Throughput = 33.28 Mbit/s
```

Very good!

mundya:
Yay C-like performance :)

mossblaser:
If you want a laugh... here's the C numbers:

```
Writing 10485760 bytes of random data to 0x60240000...
Write completed in 2463 ms! Throughput = 32.481 Mbit/s.

Reading back 10485760 bytes from 0x60240000...
Read completed in 2619 ms! Throughput = 30.546 Mbit/s.
```

...so you beat rig-scp ;) :P

mundya:
PyPy time :P

mossblaser:
> @mossblaser do we want to consider merging this now with the window size set to 1? We can investigate setting the TTO and reading the window size in an additional PR.

I would say that seems like a good idea. Will have a look over the code as it stands first though :)

mundya:
> Will have a look over the code as it stands first though :)

Of course!

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
-        packet = packets.SCPPacket(
-            reply_expected=True, tag=0xff, dest_port=0, dest_cpu=p,
-            src_port=7, src_cpu=31, dest_x=x, dest_y=y, src_x=0, src_y=0,
-            cmd_rc=cmd, seq=self._seq, arg1=arg1, arg2=arg2, arg3=arg3,
-            data=data
-        )
+        # This is implemented as a single burst packet sent using the bursty
+        # interface.  This significantly reduces code duplication.
+        # Construct a callable to retain the returned packet for us
+        class Callback(object):
```
mossblaser: Would a closure not be sufficient?

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+            bytes will be expected in a socket.
+        window_size : int
+            Number of packets which can be awaiting replies from the SpiNNaker
+            board.
+        parameters_and_callbacks: iterable of :py:class:`.scpcall`
+            Iterable of :py:class:`.scpcall` elements.  These elements can
+            specify a callback which will be called with the returned packet.
+        """
+        # Non-blocking and then the following is a busy loop.
+        self.sock.setblocking(False)
```
mossblaser: Can this not just be done just once upon socket creation?

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+
+        Parameters
+        ----------
+        buffer_size : int
+            Number of bytes held in an SCP buffer by SARK, determines how many
+            bytes will be expected in a socket.
+        window_size : int
+            Number of packets which can be awaiting replies from the SpiNNaker
+            board.
+        parameters_and_callbacks: iterable of :py:class:`.scpcall`
```
mossblaser: `s/parameters_and_callbacks/scp_packets/` ?

rig/machine_control/scp_connection.py
```diff
@@ -1,7 +1,53 @@
...
+    ----------
+    x : int
+    y : int
+    p : int
+    cmd : int
+    arg1 : int
+    arg2 : int
+    arg3 : int
+    data : bytes
+    expected_args : int
```
mundya: This is never used.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+        Parameters
+        ----------
+        buffer_size : int
+            Number of bytes held in an SCP buffer by SARK, determines how many
+            bytes will be expected in a socket.
+        window_size : int
+            Number of packets which can be awaiting replies from the SpiNNaker
+            board.
+        parameters_and_callbacks: iterable of :py:class:`.scpcall`
+            Iterable of :py:class:`.scpcall` elements.  These elements can
```
mossblaser: In the current implementation, `parameters_and_callbacks` must actually *be* an iterator. You could add:

```python
parameters_and_callbacks = iter(parameters_and_callbacks)
```

To fix.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+                        args.timeout
+                    )
 
-        # The packet we transmitted wasn't acknowledged.
-        raise TimeoutError(
-            "Exceeded {} tries when trying to transmit packet.".format(
-                self.n_tries)
-        )
+                    # Actually send the packet
+                    self.sock.send(packet.bytestring)
```
mundya: Calling `packet.bytestring` twice seems silly.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+                    raise self.error_codes[rc]
+
+                # Look up the sequence index of packet in the list of
+                # outstanding packets.  We may have already processed a
+                # response for this packet (indicating that the response was
+                # delayed and we retransmitted the initial message) in which
+                # case we can silently ignore the returned packet.
+                # XXX: There is a danger that a response was so delayed that we
+                # already reused the seq number... this is probably
+                # sufficiently unlikely that there is no problem.
```
mundya: Worth noting that the sequence values go up to 65535, so the likelihood of this occurring *is* small.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
-        raise TimeoutError(
-            "Exceeded {} tries when trying to transmit packet.".format(
-                self.n_tries)
-        )
+                    # Actually send the packet
+                    self.sock.send(packet.bytestring)
 
-    def read(self, buffer_size, x, y, p, address, length_bytes):
+            # Listen on the socket for an acknowledgement packet, there not be
+            # one.
```
mossblaser: > Listen on the socket for an acknowledgement packet, there not be one.

Yaaarrrr mee harties!

mundya: Shiver me timbers!  I wondered if you'd spot that.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+            try:
+                ack = self.sock.recv(receive_length)
```
mossblaser: Ah... a busy loop. I think I'd push for swapping this out for a select+timeout (with timeout set to the next timeout to occur) before we merge this branch in. Shouldn't be a big change (I don't think!) but should save battery-users and multi-taskers some time!

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+            except IOError:
+                # There wasn't a returned packet, we may spend quite some time
+                # here.
+                ack = None
+
+            # Process the received packet (if there is one)
+            if ack is not None:
+                # Extract the sequence number from the bytestring, iff possible
+                rc, seq = struct.unpack_from("<2H", ack,
```
mossblaser: Doesn't the format string at the start not imply the length here too?

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+                                             consts.SDP_HEADER_LENGTH + 2)
+
+                # If the code is an error then we respond immediately
+                if rc in self.error_codes:
+                    raise self.error_codes[rc]
```
mossblaser: Though this is certainly fine for the intermediate future, this behaviour does worry me somewhat and probably deserves some careful thought.

rig/machine_control/scp_connection.py
```diff
@@ -167,31 +316,42 @@ def read(self, buffer_size, x, y, p, address, length_bytes):
...
         # Prepare the buffer to receive the incoming data
         data = bytearray(length_bytes)
+        mem = memoryview(data)
+
+        # Create a callback which will write the data from a packet into a
+        # memoryview.
+        def callback(mem, data):
+            mem[:] = data[6 + consts.SDP_HEADER_LENGTH:]
+
+        # Create a generator that will generate request packets and store data
```
mossblaser: Now *this* is a great use-case for generators.

mossblaser:
Looks jolly good to me, though that excludes the busy loop(!) which should be an easy enough fix and drop the CPU loading down a fair bit. Plus would save a *load* of "get time" system calls (so actual performance may be ever-so-slightly improved... but that is a dubious claim ;)).

rig/machine_control/scp_connection.py
mundya: Yes. Why? The constant is the offset.

rig/machine_control/scp_connection.py
mundya: As in, throwing mid event loop?

mossblaser: Yeah. The end-user result is an unknown amount of things have been cancelled. In most callback architectures I've seen the convention is to have an errback with each callback.

rig/machine_control/scp_connection.py
mossblaser: > this is probably sufficiently unlikely that there is no problem.

Is that not what that says?

rig/machine_control/scp_connection.py
mundya: As noted offline, select is equivalent to setting the socket timeout in this case.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+        queued_packets = True
+        outstanding_packets = {}
+
+        # While there are packets in the queue or packets for which we are
+        # still awaiting returns then continue to loop.
+        while queued_packets or outstanding_packets:
+            # If there are fewer outstanding packets than the window can take
+            # and we still might have packets left to send then transmit a
+            # packet and add it to the list of outstanding packets.
+            if len(outstanding_packets) < window_size and queued_packets:
```
mundya: Noted by @mossblaser: this should be a `while`

rig/machine_control/scp_connection.py
mundya: Cool, I'm happy to add that in.

rig/machine_control/scp_connection.py
```diff
@@ -79,65 +125,167 @@ def send_scp(self, buffer_size, x, y, p, cmd, arg1=0, arg2=0, arg3=0,
...
+            __slots__ = ["callback", "packet", "expected_args", "n_tries",
+                         "time_sent", "extra_timeout"]
+
+            def __init__(self, callback, packet, expected_args, extra_timeout):
+                self.callback = callback
+                self.packet = packet
+                self.expected_args = expected_args
+                self.extra_timeout = extra_timeout
+                self.n_tries = 1
+                self.time_sent = time.time()
```
mossblaser: Wouldn't it be best to swap this field for "the time to timeout", that way you save yourself messing with working out the extra_timeout stuff every time you check it?

rig/machine_control/scp_connection.py
mundya: No, because the received packet has to be stored somewhere.

In Python 3 I can do:
```python
packet = None
def callback(received_bytes):
    nonlocal packet
    packet = ...
```

But not in Python 2.

rig/machine_control/scp_connection.py
mundya: I prefer parameters and callbacks because that is what's passed in, packets are created internally.

rig/machine_control/scp_connection.py
mundya: Fixed.

mundya:
I'm finding that select, and switching to using a `while` instead of an `if` make for a fairly significant performance hit.  You're welcome to play.

mundya:
Also, that 100% is not 100%... there's a branch missing in `scp_connection.py`.

mundya:
With the exception of `select`, which I graciously leave open for @mossblaser to do, I'm happy that we consider this for merging soon.  Tasks for SpiNNaker 1.4 are to investigate querying the correct window size and setting the TTO.

mundya:
@mossblaser, if I merge in your select work and get the tests to pass again can we consider merging this soon?

mossblaser:
That sounds like a plan :) thanks!

Jonathan

mundya:
I'm not sure what's up with Coveralls... coverage remains at 100% though.  @mossblaser, are we happy with this?

mossblaser:
Yep I'm happy. Silly Coveralls... Merge and release at will! Sorry about the delay ;)

mundya:
Thanks

Issue #96, opened 2015-04-17
Meet pytest's recommendations


mundya:
- [ ] Remove `__init__.py` files from `test/` directories.

mossblaser:
Mutually deemed not a big enough problem to put effort into.

PR #119, opened 2015-04-27
Test setuptools installation in test-suite. 

### LINKED ISSUE #116: Binaries not included in pip installation.

mundya:
* [x] Fix issue #116 (missing binaries in packaged distribution)
* [x] Make the test-suite/travis/tox catch such packaging issues in the future. (Demonstrated in #120)

This commit applies some relatively major file-tree restructuring to enable
pytest to run tests against an installed version of the Rig library. As a result
of this installation becomes a necessary step in testing and thus a bad
setup.py can cause tests to fail.

* Remove __init__.py from test directories.

  As per py.test's recommendations (and previously brought up in #96).

  This is the first step towards making the test suite run against an
  *installed* version of the software rather than the copy in which the test
  code runs. As a bonus, the test code nolonger gets included in the packaged
  software.

* Move tests into own directory.

  This makes the test suite run against the installed version rather than the
  version they happen to be sitting in the same module as.

* Make travis and tox install rig prior to testing.

  This tests the setup is actually working.

* Run py.test in multiple stages to work-around pytest path issues

  In order to cope with pytest's modification of the python module path, the
  test suite and doctests must be run separately for system-wide installations
  of the library-under-test...

  Additionally, the test-dependencies have been removed from setup.py since
  there is no value to them being there.

* Points test suite coverage checks at the installed version of rig

  This process is somewhat ugly but it does work and considering this allows the
  packaging to be tested too, this seems like a good thing overall.


mundya:
After review + merge please tag a new release!

mossblaser:
> I'm a little unsure why Tox didn't highlight this.

The problem doesn't manifest for development installs -- is that what tox does? I presume travis does this too. Neither make any hints in this regard in their configs though...

I'm not too sure what the solution is to this... Having the tests run against an *installed* version of Rig would probably be a good thing but I don't really know what pytest says about this. Its really not something we want to happen again...

Anyway, this fix LGTM non-the-less. I'll have a quick look at pytest's docs and see if there is any guidance in there. If there is I'll add it to this branch (bugfixes without test-fixes are bad and all that...), if not I'll merge and make an issue. Anyway, once things look good again I'll merge and release.

mossblaser:
Ah-ha! Those tests failing (with import errors) is perfect! That means after these changes the test suite *requires* that rig be installed and thus it must be testing against the installed version! Hooray!

mundya:
I've never been so happy to see failing tests...

mossblaser:
...nope same problem tox was having then... Python packages and py.test magic are a func combination. At present, py.test is adding the working directory to its path which in turn means that it finds two instances of rig (the system copy and the current copy) and pytest balks at this. (It actually runs the tests against the local version, though).

mossblaser:
More TODOs: work out how to make tox cache everything except the rig install itself to force that to happen on each run. (maybe?)

mundya:
I definitely don't recalling writing all that stuff at the top!

PR #120, opened 2015-04-28
Verify fix #116 **INTENTIONALLY BROKEN BRANCH**

### LINKED ISSUE #116: Binaries not included in pip installation.

mossblaser:
This commit breaks the Rig packaging and is intended to test that travis can
can catch such issues following the changes in #119.

THIS BRANCH SHOULD NOT BE MERGED INTO MASTER!


mossblaser:
Good! This fails. I've cancelled the rest of the tests since failing takes a while and there's no need to see it fail 6 times...

PR #121, opened 2015-04-28
Use README.rst as PyPi long description.

mundya:


setup.py
```diff
@@ -1,6 +1,57 @@
...
+                             text, re.MULTILINE):
+        orig_url = match.groupdict()["url"]
+        url = get_new_url(orig_url)
+
+        text = text.replace(".. image:: {}".format(orig_url),
+                            ".. image:: {}".format(url))
+
+    return text
+
+long_description = replace_local_hyperlinks(read_file("README.rst"))
```
mossblaser: Is it not more conventional to place this line within the setup call? (/nitpick to the XTREAM)

setup.py
```diff
@@ -1,6 +1,57 @@
...
+    encoding = kwargs.get("encoding", "utf-8")
+
+    with io.open(filename, encoding=encoding) as f:
+        return f.read()
+
+
+def replace_local_hyperlinks(
+        text, base_url="https://github.com/project-rig/rig/blob/master/"):
+    """Replace local hyperlinks in RST with absolute addresses using the given
+    base URL.
```
mossblaser: "This is used to make links in the long description function correctly outside of the repository (e.g. when published on PyPi)."

README.rst
```diff
@@ -72,7 +83,6 @@ Python Version Support
 Rig is tested against the following versions of Python:
 
 * 2.7
-* 3.3
```
mossblaser: This should really be done in another commit along with tox/travis changes -- will make a PR with this change shortly.

Issue #76, opened 2015-03-22
Proposal: Drop Python 3.3 support


mossblaser:
* Who ships python 3.3 which should be considered an important target?
* Would cut 33% of unit test runtime
* Make is far easier to reproduce tests on modern linux distributions using tox (python 3.3 isn't easy to get)

mundya:
> - Who ships python 3.3 which should be considered an important target?

The version of Fedora on my work machine... :cry: 

In principle I agree, supporting the most recent version of 3 and 2.7 should probably be sufficient.

mundya:
Is "cutting test time" a good reason to drop support for something?

mossblaser:
Lol, not really, though it is the most obvious source of developer resource drain that supporting unnecessary platforms produces...

mossblaser:
My goodness  was it only 8 days...

Anyway, this is being reopened and will be applied after #119 #121 have been merged in.

mundya:
> My goodness was it only 8 days...
March != April... sorry

mossblaser:
> mossblaser closed this 8 days ago

mundya:
|| mossblaser closed this 8 days ago

Fair enough. Time flies.

PR #122, opened 2015-04-28
Remove Python 3.3 support.

mossblaser:
As proposed in #76, official support (and thus test infrastructure) for Python
3.3 has been dropped.


mundya:
LGTM, waiting on tests.

PR #123, opened 2015-04-28
Use README.rst as PyPi long description.

mundya:


mundya:
I'll tag and release after you ok this.

mossblaser:
Assuming this approximately matches the [testpypi release](https://testpypi.python.org/pypi/rig/) I saw before I'm very happy with this.

Change log entries to my understanding are R/W and I/O performance generally, fixed packaging, tests in own directory and python 3.3 support removal.

Issue #124, opened 2015-04-29
(Optional) Buffered writing via file-IO.


mossblaser:
In my network_tester application, and potentially others, a convenient mode of operation is to pass (slices of) MemoryIO objects around to different parts of the program which may then make small writes to memory which may be to sequential regions. As such, for efficiency it would be beneficial to combine these small sequential writes into a single continuous write. This means that maximum-sized writes, and thus maximum network efficiency can be achieved and also that future introduction of #117 (Windowed reads/writes) could be made even more widely applicable.

I propose the following:

* Add a `write_buffer_size` property which defaults to 0 (meaning maintain current behaviour) but can be set to an arbitrary number of bytes (up to the size of the slice).
* When calling `write(...)`, if the write would be contiguous with previously buffered data and the data will fit in the allocated write-buffer, the data is silently buffered and not sent. Otherwise, the data is sent immediately.
* Add `flush()` method which sends any data currently in the write buffer.
* Slices of the same original MemoryIO share the same write buffer: this allows writes within slices to be merged too.

Note that:

* Only buffering contiguous writes is an intentional simplification of the problem since otherwise you have to deal with re-ordering buffered writes etc. Since the same API would be presented by a more sophisticated system, this mechanism could be implemented at a later point without breaking backward compatibility.
* Sadly, unlike normal OS buffered writes, nothing happens until you hit `flush()`. As such if you use buffered writes, you *must* hit flush at some point. Still, I think this is still useful since it allows coalescing of writes with minimal code in the user application.

@mundya @neworderofjamie Any thoughts as implementers of related systems?

mundya:
+1

neworderofjamie:
This sounds a fine idea to me - I might be getting the wrong end of the
stick here, but perhaps flushing could be automated so it always flushes
when you perform a non-coalesced write and always when you 'close' your
memory handle in some way.

mundya:
> flushing could be automated so it always flushes when you perform a non-coalesced write

This, and perhaps whenever you perform a read.

Sadly, detecting object deletion in Python is not reliable:

> so `__del__` can't serve as a replacement for good coding practices (like always closing a connection when you're done with it. **In fact, `__del__` should almost never be used because of the precarious circumstances under which it is called; use it with caution!**

http://www.rafekettler.com/magicmethods.html

But we could do the standard Python `with` syntax for flushing.

```python
with file_like:
    file_like.write(...)
# file_like flushed on leaving the `with`
```

[...]

PR #125, opened 2015-04-29
Buffered writes

mundya:
As suggested in #124.

Allows writes to be buffered.  A `flush` method and `with` manager are provided to ensure that writes are completed.

The following events flush write buffers:
 - `flush`
 - `seek`
 - `read`
 - A non-coalesced `write`

The following are equivalent:
```python
f.seek(0)
f.write(b"Hello, world")
f.flush()
```
```python
with f:
    f.seek(0)
    f.write(b"Hello, world")
```
 - [x] Stop seek from flushing
 - [x] Set buffer sizes
 - [x] Overwrite buffered values
 - [x] Writing when receiving values > buffer size
 - [x] `close`

### DIFF
--- a/docs/source/control_tutorials.rst
+++ b/docs/source/control_tutorials.rst
@@ -195,12 +195,35 @@ constrained to the allocated region. ::
     >>> block.read(13)
     b"Hello, world!"
 
-These blocks can also be sliced to allow a single allocation to be safely
-divided between different parts of the application::
+Writes can be optionally buffered before being transmitted to the SpiNNaker
+board.  This can be configured by using the `buffer_size` keyword argument::
+
+    >>> # Allocate 1024 bytes of SDRAM with tag '3' on chip (0, 0) with a
+    >>> # 256-byte write buffer
+    >>> block = mc.sdram_alloc_as_filelike(1024, 3, 0, 0, buffer_size=256)
+    >>> block.buffer_size
+    256
+
+If buffering is used then
+:py:meth:`~.rig.machine_control.machine_controller.MemoryIO.flush` must be
+called to force writes on a given file-like (and its siblings -- see below) to
+be completed::
+
+   >>> block.seek(0)
+   >>> block.write("Hello")
+   >>> block.flush()
+
+This is not necessary for unbuffered file-like objects (the default).
+
+File-like views of memory can also be sliced to allow a single allocation to be
+safely divided between different parts of the application::
 
     >>> hello = block[0:5]
     >>> hello.read()
     b"Hello"
+
+Slices of the same memory file-like are considered to be siblings and flushing
+one of them will result in flushing of the write buffer for all siblings.
 
 The :py:func:`~rig.machine_control.utils.sdram_alloc_for_vertices` utility
 function is provided to allocate multiple SDRAM blocks simultaneously.  This
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -1,6 +1,7 @@
 """A high level interface for controlling a SpiNNaker system."""
 
 import collections
+import functools
 import os
 import six
 from six import iteritems
@@ -9,15 +10,15 @@ import struct
 import time
 import pkg_resources
 
-from . import struct_file
 from .consts import SCPCommands, NNCommands, NNConstants, AppFlags, LEDAction
-from . import boot, consts, regions
+from . import boot, consts, regions, struct_file
 from .scp_connection import SCPConnection
 
 from rig import routing_table
 from rig.machine import Cores, SDRAM, SRAM, Links, Machine
 
 from rig.utils.contexts import ContextMixin, Required
+from rig.utils.docstrings import add_signature_to_docstring
 
 
 class MachineController(ContextMixin):
@@ -573,9 +574,17 @@ class MachineController(ContextMixin):
 
     @ContextMixin.use_contextual_arguments()
     def sdram_alloc_as_filelike(self, size, tag=0, x=Required, y=Required,
-                                app_id=Required):
+                                app_id=Required, buffer_size=0):
         """Like :py:meth:`.sdram_alloc` but returns a file-like object which
         allows safe reading and writing to the block that is allocated.
+
+        Other Parameters
+        ----------------
+        buffer_size : int
+            The number of bytes to store in the write buffer for the file-like.
+            If this is set to anything but `0` (the default) then
+            :py:meth:`~.MemoryIO.flush` should be called to ensure that all
+            writes are completed.
 
         Returns
         -------
@@ -592,7 +601,8 @@ class MachineController(ContextMixin):
         # Perform the malloc
         start_address = self.sdram_alloc(size, tag, x, y, app_id)
 
-        return MemoryIO(self, x, y, start_address, start_address + size)
+        return MemoryIO(self, x, y, start_address, start_address + size,
+                        buffer_size=buffer_size)
 
     def _get_next_nn_id(self):
         """Get the next nearest neighbour ID."""
@@ -1378,6 +1388,18 @@ class SpiNNakerLoadingError(Exception):
             "Failed to load applications to cores {}".format(", ".join(cores)))
 
 
+def _if_not_closed(f):
+    """Run the method iff. the memory view hasn't been closed."""
+    @add_signature_to_docstring(f)
+    @functools.wraps(f)
+    def f_(self, *args, **kwargs):
+        if self.closed:
+            raise OSError
+        return f(self, *args, **kwargs)
+
+    return f_
+
+
 class MemoryIO(object):
     """A file-like view into a subspace of the memory-space of a chip.
 
@@ -1395,7 +1417,8 @@ class MemoryIO(object):
         b"Hello"
     """
 
-    def __init__(self, machine_controller, x, y, start_address, end_address):
+    def __init__(self, machine_controller, x, y, start_address, end_address,
+                 buffer_size=0, _write_buffer=None):
         """Create a file-like view onto a subset of the memory-space of a chip.
 
         Parameters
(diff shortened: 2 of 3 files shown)


mundya:
I think this covers everything.

A new write buffer is added to `MemoryIO`.  Flushing any `MemoryIO` will flush its children, siblings and parent.  Any of the above listed events causes a flush and writes will be buffered regardless if they come before or after.

The only way I can see to screw things up is to manually create two `MemoryIO`s covering the same area of memory... but then you have a problem regardless of this PR -- and we should all be using `sdram_alloc_as_filelike` anyway.

docs/source/control_tutorials.rst
```diff
@@ -195,6 +195,25 @@ constrained to the allocated region. ::
     >>> block.read(13)
     b"Hello, world!"
 
+.. warning::
+    Writes are buffered before being transmitted to the SpiNNaker board.
+    :py:meth:`~.rig.machine_control.machine_controller.MemoryIO.flush` must be
+    called to force writes on a given file-like (and its siblings -- see below)
+    to be completed.
```
mossblaser: Is there no way to enable unbuffered mode? (Also, shouldn't that be default -- enabling buffering is something to do only when you know what you're doing and care about performance, I'd hope).

rig/machine_control/machine_controller.py
```diff
@@ -1578,6 +1611,9 @@ def seek(self, n_bytes, from_what=os.SEEK_SET):
             Note that `os.SEEK_END`, `os.SEEK_CUR` and `os.SEEK_SET` are also
             valid arguments.
         """
+        # Flush this write buffer
+        self.flush()
```
mossblaser: As mentioned before, I'd prefer seek didn't flush, that way you can always seek before writing without obliterating the buffering behaviour.

rig/machine_control/machine_controller.py
```diff
@@ -1591,6 +1627,72 @@ def seek(self, n_bytes, from_what=os.SEEK_SET):
...
+        self.parent.flush()
+
+
+class _WriteEntry(collections.namedtuple("_WriteEntry", "start, data")):
+    @property
+    def end(self):
+        return self.start + len(self.data)
+
+
+class _WriteBuffer(_WriteBufferChild):
```
mossblaser: I don't think I understand why this needs to be a tree? Why can't everyone just have a reference to the same _WriteBuffer everywhere?

mundya: You are, of course, correct.

docs/source/control_tutorials.rst
mundya: Yeah, I'll fix that. Sorry.

mossblaser:
Aside from changes proposed in #127 and buffer-size selection, LGTM :).

If buffer sizes are fixed, you can pre-allocate the buffers which should help reduce the amount of copying going on. Unfortunately this a) makes prepending expensive b) makes it difficult to enable an "unlimited size buffer".

I'd be tempted to say that prepending is not really a required feature and perhaps could be dropped in favour of the above. Further, it means you don't need to compute end-address every time... ;)

mundya:
> If buffer sizes are fixed, you can pre-allocate the buffers which should help reduce the amount of copying going on. Unfortunately this a) makes prepending expensive b) makes it difficult to enable an "unlimited size buffer".
>
> I'd be tempted to say that prepending is not really a required feature and perhaps could be dropped in favour of the above. Further, it means you don't need to compute end-address every time... ;)

I'm good with this.

mossblaser:
We should probably implement some sort of "close()" method which would probably just call flush. That way, this thing looks a lot more like a regular file, especially in write-only situations.

rig/machine_control/machine_controller.py
```diff
@@ -1578,6 +1655,9 @@ def seek(self, n_bytes, from_what=os.SEEK_SET):
             Note that `os.SEEK_END`, `os.SEEK_CUR` and `os.SEEK_SET` are also
             valid arguments.
         """
+        if self.closed:
+            raise OSError
```
mundya: Decorator time... maybe

rig/machine_control/machine_controller.py
```diff
@@ -1591,6 +1671,71 @@ def seek(self, n_bytes, from_what=os.SEEK_SET):
...
+        # A buffer of writes
+        self.buffer = bytearray(buffer_size)
+        self.written_slices = list()
+
+        self.start_address = None
+        self.buffer_size = buffer_size
+
+    def include_in_buffer(self, start, end, data):
+        self.buffer[start:end] = data
+        self.written_slices.append(slice(start, end))
```
mundya: This could just be inlined below.

mundya:
@mossblaser, @neworderofjamie your thoughts please...

mundya:
So doctests aren't included in coverage... right.

mundya:
Cool as that slice minimiser is performance on an 8MB buffer will be horrendous. I'll rip that out.

rig/machine_control/machine_controller.py
```diff
@@ -1513,9 +1576,15 @@ def read(self, n_bytes=-1):
         self._offset += n_bytes
         return data
 
+    @_if_not_closed
     def write(self, bytes):
         """Write data to the memory.
 
+        .. warning::
+            Writes are buffered before being written to the machine.
```
mundya: "Writes may be buffered"

rig/machine_control/machine_controller.py
```diff
@@ -1378,6 +1388,17 @@ def __str__(self):
             "Failed to load applications to cores {}".format(", ".join(cores)))
 
 
+def _if_not_closed(f):
+    """Run the method iff. the memory view hasn't been closed."""
+    @functools.wraps(f)
```
mossblaser: Sadly, you may well need [`add_signature_to_docstring`](https://github.com/project-rig/rig/blob/master/rig/utils/docstrings.py#L51) to make the docs come out correctly after this :(. Sigh... Python 2...

rig/machine_control/machine_controller.py
```diff
@@ -1479,6 +1529,16 @@ def __len__(self):
...
         return self._end_address - self._start_address
 
+    def __enter__(self):
+        """Enter a new block which will call :py:meth:`~.flush` when exited."""
+        return self
+
+    def __exit__(self, exception_type, exception_value, traceback):
+        """Exit a block and call :py:meth:`~.flush`."""
+        # We don't handle the exception, but we will call flush
+        self.flush()
```
mossblaser: To match the common pattern with `with` and files:

```python
with open("/some/file", "r") as f:
    # ...
```

The `__exit__` should probably close the file(?).

rig/machine_control/machine_controller.py
```diff
@@ -1427,6 +1460,22 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+        """Return the number of bytes in the write buffer."""
+        return self._write_buffer.buffer_size
+
+    def close(self):
+        """Flush and close the file-like."""
+        if not self._closed:
+            self.flush()
+            self._closed = True
+
+    @property
```
mossblaser: Why a property?

rig/machine_control/machine_controller.py
```diff
@@ -1427,6 +1460,22 @@ def __init__(self, machine_controller, x, y, start_address, end_address):
...
+    def buffer_size(self):
+        """Return the number of bytes in the write buffer."""
+        return self._write_buffer.buffer_size
+
+    def close(self):
+        """Flush and close the file-like."""
+        if not self._closed:
+            self.flush()
+            self._closed = True
```
mossblaser: `del self._write_buffer` (?)

rig/machine_control/machine_controller.py
mossblaser: "...if buffer_size is != 0"

rig/machine_control/machine_controller.py
neworderofjamie: IMHO, that's a perfect use case for a property...

mossblaser: As in for keeping it read-only? Fair enough then :).

neworderofjamie: More that it's exposing a property of a internal object that it's not
expected the programmer will have knowledge of

rig/machine_control/machine_controller.py
```diff
@@ -1591,6 +1671,80 @@ def seek(self, n_bytes, from_what=os.SEEK_SET):
...
+        # otherwise we flush the current buffer and start again
+        start_offset = start_address - self.start_address
+        end_offset = start_offset + len(data)  # Byte AFTER the end of the data
+
+        if start_offset <= self.current_end and end_offset <= self.buffer_size:
+            # The write is entirely contained within the buffer and starts
+            # within the area of the buffer which already contains data, so we
+            # can just modify the buffer.
+            self.buffer[start_offset:end_offset] = data
+            self.current_end = max(end_offset, self.current_end)
```
mossblaser: I think the fact that not every write made will actually get to the hardware (when buffering is enabled) probably warrants documenting, though I can't really see it ever biting someone... if you care about exactly what writes are happening you should be flushing accordingly...

tests/machine_control/test_machine_controller.py
```diff
@@ -300,6 +300,7 @@ def test_sdram_alloc_as_filelike_read_write(self, controller, data):
         with controller(x=1, y=0):
             mem = controller.sdram_alloc_as_filelike(len(data))
             assert mem.write(data) == len(data)
+            mem.flush()
```
mossblaser: Surely this shouldn't be required here since buffer_size=0?

mossblaser:
Aside from super-high nitpicking,

Rigged

neworderofjamie:
Agreed! Especially as I thought my SDRAM as a file-like thing hobby horse
had been abandoned!

rig/machine_control/machine_controller.py
mossblaser: Aha! @mundya has just pointed out to me that we might be talking about different things: I was talking about the `closed` property. I presume you are talking about the `buffer_size` property? The GitHub line comment stuff is displayed a bit confusingly in summaries...

neworderofjamie: Yes, that makes more sense - I would personally make that a property too to
make it's readonlyness explicit, but I realise that possibly betrays my C++
fear of public members variables...

mossblaser: <dramatic action film trailer voice> ...in the land of the Python, nobody is safe! <.../>

rig/machine_control/machine_controller.py
mundya: This is done in the `write` documentation (1579 onward).

PR #127, opened 2015-04-30
Simplify buffer implementation.

mossblaser:
As suggested by my own comment in #125 .

**Note: this PR merges into the coalesced-writes branch.**


mundya:
Should have been like this originally... thank you.

PR #137, opened 2015-05-10
Clearer error message on SDRAM allocation failure

mundya:
The error message returned when allocating SDRAM fails isn't clear that
the cause can be either that there is insufficient memory OR that the
tag assigned to the allocation is already in use.

This is now changed so that when insufficient memory is definitely the
cause (only when tag == 0) the tag is not mentioned, but that the tag is
mentioned when it may be the problem.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -639,7 +639,7 @@ class MachineController(ContextMixin):
         rv = self._send_scp(x, y, 0, SCPCommands.alloc_free, arg1, size, tag)
         if rv.arg1 == 0:
             # Allocation failed
-            raise SpiNNakerMemoryError(size, x, y)
+            raise SpiNNakerMemoryError(size, x, y, tag)
         return rv.arg1
 
     @ContextMixin.use_contextual_arguments()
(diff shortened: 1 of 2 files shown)


mundya:
For completeness sake this _could_ look in the `sv` struct to determine if the tag is taken... this seems a little too heavy maybe?

mossblaser:
Excellent! I've been bitten by this far too many times (especially due to SCAMP not cleaning up without apps being loaded, as is often the case when my script fails after memory allocation but before loading)...

Does the RC not indicate the exact cause of the error? If not, a feature request for ST and your supplied fix will do very nicely in the meantime!

mundya:
> especially due to SCAMP not cleaning up without apps being loaded

This is a bug; particularly with regarding to routes still being valid if they passed through chips with no executables on them.  I possibly convinced ST of the case, so hopefully 1.4 will fix that.

> Does the RC not indicate the exact cause of the error? If not, a feature request for ST and your supplied fix will do very nicely in the meantime!

No, you just get a zero back from the malloc (which is then stuck in the returned SCP packet).

mossblaser:
> This is a bug; particularly with regarding to routes still being valid if they passed through chips with no executables on them. I possibly convinced ST of the case, so hopefully 1.4 will fix that.

Definately.

> No, you just get a zero back from the malloc (which is then stuck in the returned SCP packet).

Sad times :(. I think this would be a fair situation for using arg2 to produce extra metadata since there are two usefully distinct causes of failure... Will petition ST tomorrow if I remember...

In the meantime! LGTM! :)

mundya:
> I think this would be a fair situation for using arg2 to produce extra metadata since there are two usefully distinct causes of failure

Agreed, I think it's just a case of getting the code in `cmd_alloc` to look through the tag table.

Issue #139, opened 2015-05-12
Version number in documentation is wrong


mossblaser:
This isn't kept in sync automatically... Joy!

mundya:
Do we want to go for `__version__` somewhere?

mossblaser:
I think so! I've made a "version.py" which defines `__version__` and is imported in the relevent places, setup.py just execs that file directly and the docs import `spinner.__version__`. Seems to work nicely enough!

https://github.com/SpiNNakerManchester/SpiNNer/commit/55542138e02422f5add9f42c24554c5591ac42aa

mundya:
Cool, thank you :)

PR #140, opened 2015-05-13
Make version numbers consistent throughout.

mossblaser:
In response to #139

* Defines `rig.__version__`
* Documentation and setup.py read version from the above (directly and indirectly, respectively).
* Adds --version/-V to command line utilities.


mundya:
LGTM.  Merge when ready.

PR #143, opened 2015-05-22
Merge entries with equivalent keys/masks

mundya:
This fixes a bug where entries with equivalent masks and keys were not merged. Routing table entries are now not constructed until the last possible minute, just after default entries are removed.

### DIFF
--- a/rig/place_and_route/utils.py
+++ b/rig/place_and_route/utils.py
@@ -2,7 +2,7 @@
 structures from the products of placement, allocation and routing.
 """
 
-from collections import defaultdict, deque
+from collections import defaultdict, deque, namedtuple, OrderedDict
 
 from six import iteritems
 
@@ -51,7 +51,8 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
     """Convert a set of RoutingTrees into a per-chip set of routing tables.
 
     This command produces routing tables with entries optionally omitted when
-    the route does not change direction.
+    the route does not change direction. Entries with identical keys and masks
+    will be merged.
 
     Note: The routing trees provided are assumed to be correct and continuous
     (not missing any hops). If this is not the case, the output is undefined.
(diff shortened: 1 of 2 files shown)


rig/place_and_route/utils.py
```diff
@@ -73,8 +74,11 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
     -------
     {(x, y): [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
     """
-    # {(x, y): [RoutingTableEntry, ...]
-    routing_tables = defaultdict(list)
+    # Pairs of inbound and outbound routes.
+    _InOutPair = namedtuple("_InOutPair", "ins, outs")
+
+    # (x, y) -> (key, mask) -> {_InOutPair, ...}
```
mossblaser: I don't understand this syntax(?)

rig/place_and_route/utils.py
```diff
@@ -102,8 +106,29 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+                    set([direction]), set(out_directions)
+                )
+
+    # Construct the routing tables from the route sets
+    routing_tables = defaultdict(list)
+    for (x, y), routes in iteritems(route_sets):
+        for (key, mask), route in iteritems(routes):
+            # Remove default routes where possible
+            if omit_default_routes and (len(route.ins) == 1 and
+                                        route.outs == route.ins):
```
mossblaser: Shouldn't this check that the output route is *opposite* the input route, i.e. not the same(?)

rig/place_and_route/utils.py
mundya: Sorry, "{(x, y): {(key, mask): [_InOutPair, ...], ...}, ...}"... Haskell leeches into my thinking sometimes...

rig/place_and_route/utils.py
mundya: `ins` comes from `direction` (line 91 in this commit) which appears to be the direction out of the previous node.  (I copied you here).

mossblaser: Ah, yes! Jolly good! (It would have worried me more if the tests hadn't caught that if it was a bug ;)).

mossblaser:
LGTM :) Thanks for that @mundya!

PR #145, opened 2015-05-23
Correct core mask in application loading

### LINKED ISSUE #144: Can't load application to cores 14..17
In `_send_ffe` the cores are masked by `0x3fff`. This should be `0x3ffff`.

Thanks, and apologies, to @jgosmann for running something big enough to find this.

mundya:
Applications previously couldn't be loaded onto cores 14..17 because the core mask was masked with 4 fewer bits than it needed.

Fixes #144.

@neworderofjamie can you check this seems sane please?  I think Jonathan is on holiday.


mossblaser:
Whoopsie! Fix looks good to me! Merge at will once travis is happy!

mundya:
Thanks, @mossblaser

PR #146, opened 2015-05-24
Simulated annealing based placement

mossblaser:
This PR adds a simulated-annealing based placement algorithm which very closely mimics the techniques outlined by "VPR: A New Packing, Placement and Routing Tool for FPGA Research" by Vaughn Betz and Jonathan Rose from the "1997 International Workshop on Field Programmable Logic and Applications" ([PDF](http://www.eecg.toronto.edu/~vaughn/papers/fpl97.pdf)).

The placer should fulfill all the requirements of the Rig placer prototype and thus should be ready to drop into new applications.

Performance
----------------

Though this placer is implemented in relatively inefficient Python, it is probably not unusably slow for SpiNNaker systems up to perhaps the low tens of boards. It is, however, currently the first placer which attempts to produce "good" placements. I aim to perform some runtime and quality benchmarks in the immediate future to back up these hand-wavey claims... For now, here are some ballpark numbers:

* For a **SpiNN-3 board** sized system, a full place-and-route run is **1 - 4 seconds** on my laptop for networks with 64 nodes and several multicast nets per node, depending on connectivity.

* For a **SpiNN-5 board** sized system, a 784-node pipeline-style network (stages of 10 nodes each connected all-to-all to the next stage) with 10% random connectivity is placed-and-routed in **1 minute and 32 seconds** on my laptop. In terms of quality, total net length after routing is 10,768 hops with the maximum and mean number of nets passing through a single chip is 89 and 59.5 respectively. For the hilbert placer (which runs in 1 second...) total net length is 13,417 hops and max/mean nets through a chip is 279 and 117.2 respectively. (i.e. the SA placer yields substantial improvement).

* For a **three-board** system, a 2401-node network connected as a 49x49 torus (with each node having a net connecting to its 8 neighbors) and 5% random connectivity, place-and-route completes in **9 minutes 29 seconds**. Total net length, max/mean nets-per-chip are 28,818 hops, 97 and 64.5 respectively. By comparison, for the hilbert placer (which takes 6 seconds), 46,745 hops, 257 and 189.0 come out (i.e. considerably worse).
  For a visual comparison of the respective placement solutions take a look at:
  Placement for SA vs Hilbert
  (Left: simulated annealing, Right: Hilbert). It is worth noting that the lines shown show "as-the-crow-flies" routes and not the actual router output used to generate the above stats. Also note that the very long side-to-side connections in the left image would actually be very short wrap-around hops since a three-board system is a torus.

Current (known) limitations
---------------------------------

* **The optimisation function is path length, not congestion.** Path length estimates are much easier to quickly compute and will generally indirectly improve congestion anyway.
* **The placer runs about twice as slow for systems with torus topologies.** This is due the the increased cost of computing the (minimal) bounding box around a net in a torus.
* **Heuristic cost function poorly matches high fan-out net costs.** At present the Half-Perimeter-Wire-Length (HPWL) heuristic is used (i.e. net bounding box width + height) which is known to be a poor estimator of wire length for net fan-outs > 3. There are a whole slew of alternatives in this space however I need to do some research to see how these cope with the fact that a high fan-out net may only actually fan-out to a small number of unique locations. (In real circuits, things don't get collected up into cores).
* **Heuristic cost function assumes a non-hexagonal grid.** I need to do some reading about how best to deal with this. Bounding-hexagons are harder than bounding boxes...
* **Heuristic cost function fully-evaluated for every affected net on every swap.** This means bounding-boxes are fully recomputed for all involved nets for every swap. This is possibly wasteful however quick experiments with attempting to minimise the recomputation required did not produce significantly different runtime. This would be a priority for performance improvements since the bounding-box calculation function is by far the hottest in the algorithm.
* **The placer runs much slower than it could when far more system resource is available than required.** This is because the placer has far more space to search. I may make the placer artificially constrain the search space when more resource is available than really required in a future revision.
* **The placer is written in Python...** Just to reiterate, it is *zlowwww*. Some effort has been made to optimise the implementation of the hotter loops (e.g. a 3-4x speedup is gained by not using `min` and `max`... seriously...) but the long-term plan is to write a C-based implementation. This version, however, I expect to be sufficient for casual single-board use for a surprisingly long time.

### DIFF
--- a/docs/source/place_and_route/placement_algorithms.rst
+++ b/docs/source/place_and_route/placement_algorithms.rst
@@ -1,4 +1,6 @@
 Placement algorithms
 --------------------
 
+.. autofunction:: rig.place_and_route.place.sa.place
+
 .. autofunction:: rig.place_and_route.place.hilbert.place
--- a/rig/machine.py
+++ b/rig/machine.py
@@ -100,6 +100,12 @@ class Links(IntEnum):
     def to_vector(self):
         """Given a link direction, return the equivalent vector."""
         return _direction_link_lookup[self]
+    
+    
+    @property
+    def opposite(self):
+        """Get the opposite link to the one given."""
+        return Links((self + 3) % 6)
 
 
 _link_direction_lookup = {
--- a/rig/machine.py
+++ b/rig/machine.py
@@ -203,6 +203,18 @@ class Machine(object):
             self.chip_resources, self.chip_resource_exceptions,
             self.dead_chips, self.dead_links)
 
+    def __eq__(self, other):
+        """Test whether this Machine describes the same machine as another."""
+        return (self.width == other.width and
+                self.height == other.height and
+                self.chip_resources == other.chip_resources and
+                all(self[chip] == other[chip]
+                    for chip in self.chip_resource_exceptions) and
+                all(self[chip] == other[chip]
+                    for chip in other.chip_resource_exceptions) and
+                self.dead_chips == other.dead_chips and
+                self.dead_links == other.dead_links)
+
     def __contains__(self, chip_or_link):
         """Test if a given chip or link is present and alive.
 
--- a/rig/machine.py
+++ b/rig/machine.py
@@ -100,8 +100,7 @@ class Links(IntEnum):
     def to_vector(self):
         """Given a link direction, return the equivalent vector."""
         return _direction_link_lookup[self]
-    
-    
+
     @property
     def opposite(self):
         """Get the opposite link to the one given."""
--- a/rig/netlist.py
+++ b/rig/netlist.py
@@ -41,3 +41,9 @@ class Net(object):
     def __contains__(self, vertex):
         """Test if a supplied vertex is a source or sink of this net."""
         return vertex == self.source or vertex in self.sinks
+    
+    def __iter__(self):
+        """Iterate over all vertices in the net, starting with the source."""
+        yield self.source
+        for vertex in self.sinks:
+            yield vertex
--- a/rig/netlist.py
+++ b/rig/netlist.py
@@ -41,7 +41,7 @@ class Net(object):
     def __contains__(self, vertex):
         """Test if a supplied vertex is a source or sink of this net."""
         return vertex == self.source or vertex in self.sinks
-    
+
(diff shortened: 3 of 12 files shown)


mossblaser:
@mundya Not entirely sure what to do re: code review in this instance. It would certainly be unfair to get you reviewing this in any great depth since it is nominally part of my own playing around, that said it is of course relatively important for Rig that it works...

What would you feel about making this the default placer in Rig? Its obviously a lot slower but it (should be) a lot better...

Whatever goes on, I think I'd like to postpone a merge until I've got it limiting placements for cases where the machine is much larger than the problem.

mundya:
> Not entirely sure what to do re: code review in this instance. It would certainly be unfair to get you reviewing this in any great depth since it is nominally part of my own playing around, that said it is of course relatively important for Rig that it works...

I'd like to learn (I've already had a brief glance). My suggestion would be to review it together on Thursday/Friday if you're around, or via Skype when I'm in Canada?  I'm guessing there's no rush given your last point.

> What would you feel about making this the default placer in Rig? Its obviously a lot slower but it (should be) a lot better...

I'm happy with this, if you're doing anything trivial then it probably won't be too noticeably slower and if you're doing anything serious then most likely you care about the goodness of the result. In other cases you should probably be specifying the algorithm anyway!

mossblaser:
> I'd like to learn (I've already had a brief glance). My suggestion would be to review it together on Thursday/Friday if you're around, or via Skype when I'm in Canada? I'm guessing there's no rush given your last point.

Thursday/Friday sounds good (assuming I make it in -- I plan to!).

Incidentally I'd forgotten about Canada. Perhaps it would be a good idea to try and get it merged in before then on the basis that if people use it it could make life easier for anyone that uses it if there are some spinn5 boards around like last year.

mundya:
> Perhaps it would be a good idea to try and get it merged in before then on the basis that if people use it it could make life easier for anyone that uses it if there are some spinn5 boards around like last year.

Worth noting that I arrive in Waterloo a week before the summer school starts, so there's definitely some grace time, but reviewing this in person would be nice :)

mossblaser:
According to @mundya, this looks good so I'll merge in now.

PR #147, opened 2015-05-24
Use core count to speed application loading

mundya:
The current implementation of `load_application` queries the state of each loaded core to ensure that loading was successful. This is good for cases where the application map isn't _all_ the applications that will be loaded (i.e., a count of all cores in the `wait` state is not necessarily a count of all the cores we just tried to load), or when a failure mode may allow the right number of cores in the wrong place.
However, in the majority of cases this is overkill because:

1. This failure mode isn't currently possible.
2. Most of the time `load_application` is called once to load all applications.

The result is that loading applications takes noticeably more time than loading application data.

This commit provides a new parameter to `load_application`. If `all_cores` is True then the core count is used to quickly determine whether all applications are loaded. If False (the default) then the
current method is used.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -852,19 +852,18 @@ class MachineController(ContextMixin):
             Time to pause (in seconds) after loading to ensure that the
             application successfully reaches the wait state before checking for
             success.
-        all_cores : bool
-            If True then the targets dictionary will be assumed to represent
-            _all_ the cores that will be loaded and a faster method to
-            determine whether all applications have been loaded correctly will
-            be used. If False (the default) then a fallback method will be
-            used.
+        use_count : bool
+            If True (the default) then the targets dictionary will be assumed
+            to represent _all_ the cores that will be loaded and a faster
+            method to determine whether all applications have been loaded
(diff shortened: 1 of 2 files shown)


mossblaser:
This seems good to me!

Is there any reason why this mode shouldn't be the default? It seems that the cases where this applies is likely the most common case.

Also, I think the parameter name `all_cores` could do with being clearer. (e.g. `just_count` or something).

mundya:
> Is there any reason why this mode shouldn't be the default? It seems that the cases where this applies is likely the most common case.

I was unsure on this, I'm happy to make it the default.  _Technically_ that's a backwards incompatible change...

> Also, I think the parameter name all_cores could do with being clearer. (e.g. just_count or something).

```use_count```?

mossblaser:
> I was unsure on this, I'm happy to make it the default. Technically that's a backwards incompatible change...

Maybe... Indeed, I suppose in the future the whole mechanism may well get fixed anyway to make that optimisation unnecessary. That said, pre-release stuff and all...

> `use_count`?

Sure :)

mossblaser:
LGTM! Merge when travis is happy :).

PR #148, opened 2015-05-27
Count cores in many states

mundya:
Modifies `count_cores_in_state` to accept an iterable of states to count and return the sum. This is useful, for example, in ensuring that no cores are in bad states.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -969,13 +969,22 @@ class MachineController(ContextMixin):
 
         Parameters
         ----------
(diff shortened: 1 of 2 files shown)


mossblaser:
LGTM!

mundya:
Only just now did I realise that splitting on commas would be nice...

```python
count_cores_in_state("dead, idle, exit")
```

mossblaser:
/me always feels a bit funny about this particular Python
pseudo-convention... I'd prefer to leave it :)

mundya:
Fair enough, thanks!

PR #150, opened 2015-05-28
Remove routing table entries with an stop signal

mundya:
- [x] Add a hardware test for `clear_routing_table_entries`

e223add should be removed when SpiNNaker tools 1.4 is released.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -1195,6 +1195,14 @@ class MachineController(ContextMixin):
         return table
 
     @ContextMixin.use_contextual_arguments()
+    def clear_routing_table_entries(self, x, y, app_id):
+        """Clear the routing table entries associated with a given application.
+        """
+        # Construct the arguments
+        arg1 = (app_id << 8) | consts.AllocOperations.free_rtr_by_app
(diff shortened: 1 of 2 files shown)


rig/machine_control/machine_controller.py
```diff
@@ -969,6 +969,20 @@ def send_signal(self, signal, app_id):
                 "send_signal: Cannot transmit signal of type {}".format(
                     repr(signal)))
 
+        # XXX If the signal is "stop", then we first remove all routing table
+        # entries associated with the application ID. This code will be
+        # unnecessaru when SpiNNaker tools 1.4 becomes available.
```
mossblaser: unnecessaru

mundya:
Note, the hardware test is that the routing tables are correctly removed from a chip which doesn't have any applications loaded to it.

rig/machine_control/machine_controller.py
mundya: Totallu

mossblaser:
Sounds good! Awaiting the travis-ty and then you're go to merge :)

mossblaser:
Incidentally, an issue to revert this patch on 1.4 being released should be made...

mundya:
It's listed on the general 1.4 list.

PR #154, opened 2015-06-01
Make encoding routes in RoutingTrees non-lossy.

mossblaser:
Previously, the `RoutingTree` data structure did not explicitly encode the links to use to get between each hop of the route. In some circumstances (e.g. NxM with N or M <= 2), this omission means that the structure ambiguously expresses the route produced by the routing algorithm.

The specific direction chosen by a router may be more important in future routing algorithms and, in the present implementation, makes generating accurate illustrations of a generated routing solution impossible in the general case.

This commit (backward-incompatibly) modifies the `RoutingTree` data structure to include the above information as well as generally make it a little cleaner. As far as I know, nobody (except me) is doing anything but pass this structure into the `build_routing_tables` utility method (which has been updated accordingly in this commit) and thus this breaking change shouldn't impact anyone.

This change should not backward-incompatibly change the rest of the place-and-route infrastructure.

### DIFF
--- a/rig/place_and_route/place/hilbert.py
+++ b/rig/place_and_route/place/hilbert.py
@@ -117,7 +117,8 @@ def place(vertices_resources, nets, machine, constraints):
             # Place the vertex
             unplaced_vertices.remove(constraint.vertex)
             placements[constraint.vertex] = loc
-        elif isinstance(constraint, ReserveResourceConstraint):
+        elif isinstance(constraint,  # pragma: no branch
+                        ReserveResourceConstraint):
             apply_reserve_resource_constraint(machine, constraint)
 
     # Allocate chips along a Hilbert curve large enough to cover the whole
--- a/docs/source/machine.rst
+++ b/docs/source/machine.rst
@@ -3,3 +3,4 @@
 
 .. automodule:: rig.machine
     :members:
+    :special-members:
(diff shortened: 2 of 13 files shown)


mossblaser:
Ugh... that accidental tag has thoroughly confused travis...

rig/place_and_route/place/hilbert.py
```diff
@@ -117,7 +117,8 @@ def place(vertices_resources, nets, machine, constraints):
             # Place the vertex
             unplaced_vertices.remove(constraint.vertex)
             placements[constraint.vertex] = loc
-        elif isinstance(constraint, ReserveResourceConstraint):
+        elif isinstance(constraint,  # pragma: no branch
```
mundya: Is there a reason this is difficult to test?

mossblaser: @mundya Hello from over the pond!

It isn't hard but it is not worth-while: The branch coverage in this case is complaining that it doesn't see any examples where there are constraints but the last one listed isn't a `ReserveResourceConstraint`. As such you could add a test with the constraints in an alternative order but it wouldn't really be worthwhile...

mundya: Hello!

That sounds fine, sorry!

mossblaser: That's fine! Its an annoying thing about branch coverage :/ if you have nested branchables which all end at once, there are a huge number of possible branches to cover and it is rare that all of them are meaningfully different.

mundya:
This LGTM, sorry for the delay.

mossblaser:
Thanks for that! I wasn't expecting to see much forward progress since you really do have better things to be doing the next few weeks! ;)

Issue #45, opened 2015-03-10
Router status register reading


mossblaser:
* [ ] Counters
* [ ] Histograms

PR #158, opened 2015-06-04
Add `get_router_diagnostics`

mundya:
This returns the router counter diagnostics as a namedtuple.


mossblaser:
Thank you very much for adding that one! I've just added a live test simply out of sheer paranoia since we've been bitten by it in the past ;). That said, "shockingly", it does in fact seem to work!

When you're happy with my additional test, merge at will!

I'd say this definitely covers part one of #45. I should ask Genting about his histogram-related discoveries at some point...

mundya:
That's a very slick live test, thank you. I'd really like a script that displays a heat map of dropped multicast packets, but that's for another day. Merging!

PR #159, opened 2015-06-05
Set build timestamp time-zone to UTC.

### LINKED ISSUE #157: Tests do not work in Canada
I think the BMP tests have a hard-coded timestamp test.  This works in the UK apparently, but in Canada the 1970 that is expected becomes 1969. Oops.

mossblaser:
Fixes #157.

Though I strongly suspect that the build times are actually relative to the Europe/London timezone (i.e. what ST has on his machine), it really should be UTC. Further, the error of a maximum of 1-ish hour is unlikely to cause confusion since software builds are infrequent.

A very good spot by @mundya and proof that I am yet another victim of the many [Falsehoods programmers believe about time](http://infiniteundo.com/post/25326999628/falsehoods-programmers-believe-about-time). Specifically this falls afoul of 8. "The machine that a program runs on will always be in the GMT time zone."...

### DIFF
--- a/rig/scripts/rig_info.py
+++ b/rig/scripts/rig_info.py
@@ -10,6 +10,7 @@ from six import iteritems, itervalues
 from collections import defaultdict
 
 from datetime import datetime
+from pytz import utc
 
 import rig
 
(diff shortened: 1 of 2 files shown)


mundya:
I was wracking my brains on why Travis didn't fail on this before, but because they're in Germany finding this required one of us to go west... Yay!

LGTM!

PR #160, opened 2015-06-05
Fix live get_machine test

mossblaser:
Removes an erroneous test that checked that chips with dead links had no resource exceptions (which was there due to a copy-pasta fail...).

Many thanks to @mundya for finding this bug.

(Sorry!)


mundya:
Seems fair enough! Thank you!

PR #163, opened 2015-06-12
Fix infinite loop in NER routing algorithm.

mossblaser:
In the initial route-generating code, the NER routing algorithm had a bug meant
that new routes could be inserted into a routing tree which overlapped existing
tree segments sometimes resulting in cycles being added to the routing tree.
These cycles would then cause later stages of the algorithm to get stuck in an
infinite loop.

To resolve the bug, new route segments are truncated at the first point they
come into contact with the rest of the routing tree. A fuzzing test designed to
maximise the chances of similar bugs being manifested has been added. This test
reliably finds the above bug when the fix is reverted.

Note that though this bug is severe it requires very peculiar nets to manifest
itself. In particular to make the bug likely to manifest, nets must be extremely
high fan-out (but *not* connect to all chips) and the pattern must be such that
NER ends up inserting highly serpentine routes. As such most applications are
likely not to encounter this bug in normal operation. That said, @mundya and
@neworderofjamie should ideally be aware of its existence and make sure to
update their local Rig installation when this fix is merged.

This commit also fixes a test which failed to account for some legal hops in
routes (found by the fuzzing test).

### DIFF
--- a/rig/place_and_route/route/ner.py
+++ b/rig/place_and_route/route/ner.py
@@ -76,12 +76,12 @@ def ner_net(source, destinations, width, height, wrap_around=False, radius=10):
         neighbour = None
 
         # Try to find nodes nearby by searching an enlarging concentric ring of
-        # ndoes.
+        # nodes.
         for x, y in concentric_hexagons(radius, destination):
             if wrap_around:
                 x %= width
                 y %= height
-            if (x, y) in route and (x, y) != destination:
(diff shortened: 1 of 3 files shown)


mundya:
LGTM! Feel free to merge, bump and release away!

mossblaser:
Thanks! A very embarrassing bug...

PR #164, opened 2015-06-16
Remove __slots__ from user-facing modules.

mossblaser:
The major motivation for this change is that in Python 2 you cannot pickle
objects with `__slots__` in the general case. This is problematic if you wish to
dump place/route artefacts, e.g. for the purposes of plotting.

Since I don't believe we're currently memory-bound adding `__slots__` probably
isn't very useful and so removing this seems like a good idea.

### DIFF
--- a/rig/machine.py
+++ b/rig/machine.py
@@ -173,8 +173,6 @@ class Machine(object):
         that links have two directions and both should be defined if a link is
         dead in both directions (the typical case).
     """
-    __slots__ = ["width", "height", "chip_resources",
-                 "chip_resource_exceptions", "dead_chips", "dead_links"]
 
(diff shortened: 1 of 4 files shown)


mundya:
LGTM, but is there a reason you'd rather not add the appropriate magic methods? I can see there being a lot of nets and using less memory there still seems like a good idea.

mossblaser:
I'd err on the side of "premature optimisation is bad" since if they dominate performance, we're seriously winning (and who wants to test magic methods?).

mundya:
I agree that premature optimisation is bad. I guess we can always throw things back in. Feel free to merge.

mundya:
Please bump + release :)

Issue #162, opened 2015-06-12
Find-a-Board Wizard Generator


mossblaser:
Based on the experience of both the HPB team, myself and @mundya at the Nengo Summer school, finding out all the details of your newly-connected SpiNNaker board is intimidating and confusing.

I propose including a utility function in Rig which provides the basis for an interactive wizard to try and work out what board is available.

The following information should be (optionally) obtainable from the tool:

* [x] **What is the IP address of the board?** This may be attainable from locally-connected boards by listening for the UDP broadcast packets that un-booted boards produce. This is not robust across networks and may/may not conflict with common firewall settings -- this needs a little light investigation.
* [ ] **What type of board(s) does the user have?** This one is easy: just ask a multiple-choice question.
* [x] **What is the dimensionality of the network?** If the user has a single board, this is easy. If the user has multiple boards, a sensible guess can be made by asking how many boards they have.
* [ ] **What are the BMP IPs?** This is unimportant for all but multiboard systems where it may be necessary to power-on the boards. (I'll probably leave this for another day... This along with booting in applications could do with a bit of thought...)

The plan would be to make a generator which generates questions and accepts answers. This generator could potentially be integrated into some fancy GUI fairly easily. A straight-forward CLI wrapper function will also be provided which can be dropped in to application-specific CLI applications.

mossblaser:
I think the changes in #165 are sufficient for now and the framework is there for adding the additional features above as needed.

PR #165, opened 2015-06-16
Tools for helping users get set up with SpiNNaker

mossblaser:
* Adds `rig-discover` to discover the IP of locally connected, unbooted SpiNNaker boards. This uses a new function in `rig.machine_controller.unbooted_ping` to implement this functionality.
* Adds wizards which ask simple questions which can coax information about SpiNNaker systems from novice users.

Implements the more major parts of #162. I'll try and build something suitable for Nengo config file setup based on this...

**Note: Travis will fail for the moment while the SpiNNaker proxy is down. Tests do pass locally.**

### DIFF
--- a/tests/test_geometry.py
+++ b/tests/test_geometry.py
@@ -1,6 +1,9 @@
+import pytest
+
 from rig.geometry import concentric_hexagons, to_xyz, minimise_xyz, \
     shortest_mesh_path_length, shortest_mesh_path, \
-    shortest_torus_path_length, shortest_torus_path
+    shortest_torus_path_length, shortest_torus_path, \
+    standard_system_dimensions
 
 
 def test_concentric_hexagons():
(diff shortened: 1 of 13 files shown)


tests/test_geometry.py
```diff
@@ -309,3 +312,25 @@ def test_shortest_torus_path():
...
                 (start, end, width, height, neg_minimiseds)
+
+
+def test_standard_system_dimensions():
+    # Special case: 0
+    assert standard_system_dimensions(0) == (0, 0)
+
+    # Should crash on non-multiples of 3
+    with pytest.raises(ValueError):
+        standard_system_dimensions(1)
```
mundya: This should be special cased, right? I can legitimately have a 1 board system.

mossblaser: Maybe... Though a one board system doesn't really have 8x8 chips ;). That said it seems like a reasonable expectation to have! Will add :)

mundya:
This seems very cool! Feel free to merge at will post type corrections!

PR #167, opened 2015-06-19
Fix Numpy 1.8.1 issue

mundya:
Calling `np.uint32([[1.0]])` returns a 0-d array.  This avoids this type
of call and should retain the shape of the original array when
typecasting.

Could someone with NumPy 1.8.1 confirm that the added test fails without the change to `rig/type_casts.py`? @neworderofjamie?

(Reported by @tcstewar)

### DIFF
--- a/rig/type_casts.py
+++ b/rig/type_casts.py
@@ -240,7 +240,7 @@ class NumpyFloatToFixConverter(object):
         vals *= 2.0 ** self.n_frac
         vals = np.round(vals)
 
-        return self.dtype(vals)
+        return np.array(vals, dtype=self.dtype)
 
 
 class NumpyFixToFloatConverter(object):
--- a/tests/test_type_casts.py
+++ b/tests/test_type_casts.py
@@ -199,6 +199,17 @@ class TestNumpyFloatToFixConverter(object):
                         *[ftf(v) for v in values])
         )
 
+    def test_shape_is_retained(self):
+        # Format
+        fpf = NumpyFloatToFixConverter(True, 32, 15)
+
+        # Create the input matrix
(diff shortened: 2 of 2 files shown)


neworderofjamie:
Works for me!

mossblaser:
With numpy 1.8.1 in a virtualenv with Python 3 on Arch Linux the new test does *not* fail when the change to type_casts.py is reverted.

    $ virtualenv .
    ...
    $ . bin/activate
    $ pip install numpy===1.8.1
    ...
    $ python
    Python 3.4.3 (default, Mar 25 2015, 17:13:50) 
    [GCC 4.9.2 20150304 (prerelease)] on linux
    Type "help", "copyright", "credits" or "license" for more information.
    >>> import numpy
    >>> numpy.__version__
    '1.8.1'
    [git clone and grab the typecasts file from master]
    $ py.test tests/test_type_casts.py 
    ==================================================================== test session starts ================
    platform linux -- Python 3.4.3 -- py-1.4.28 -- pytest-2.7.1
    rootdir: /tmp/rig, inifile: pytest.ini
    plugins: cov
    collected 87 items 
    
    tests/test_type_casts.py ................................................................................
    
    ================================================================= 87 passed in 0.19 seconds =============

mundya:
> With numpy 1.8.1 in a virtualenv with Python 3 on Arch Linux the new test does not fail when the change to type_casts.py is reverted.

Damn... Could be a Windows NumPy thing? @tcstewar, can you try running the test without the fix and see what happens?

mundya:
It appears that we redid this PR as #179.

mossblaser:
Ugh... Nicely done anyway! :)

PR #168, opened 2015-06-20
Speed up application loading

mundya:
This commit uses `send_scp_burst` to accelerate the loading of application executables to SpiNNaker.  As they are always flood-filled from (0, 0, 0) this should result in a factor 8 speed-up of copying the application across when windowing is enabled (from 4Mbit/s to 32Mbit/s).

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -12,7 +12,7 @@ import pkg_resources
 
 from .consts import SCPCommands, NNCommands, NNConstants, AppFlags, LEDAction
 from . import boot, consts, regions, struct_file
-from .scp_connection import SCPConnection
+from .scp_connection import SCPConnection, scpcall
 
 from rig import routing_table
 from rig.machine import Cores, SDRAM, SRAM, Links, Machine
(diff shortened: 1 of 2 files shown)


rig/machine_control/machine_controller.py
```diff
@@ -719,26 +719,35 @@ def _send_ffs(self, pid, region, n_blocks, fr):
...
-            address += data_size
-            pos += data_size
+
+            # Store the current block number and offset
+            block = 0
+            pos = 0
+            while pos < aplx_size:
+                # Get the next block of data, send and progress the block
+                # counter and the address
+                data = aplx_data[pos:pos + self.scp_data_length]
```
mossblaser: Is it possible to do some byte-array magic here to avoid copying?

mossblaser:
Looks sensible enough to me! That said I'd look forward to the prospect of having additive flood-fill regions making this optimisation relatively unimportant. Incidentally, do we know what the effect is of upping the loading throughput on flood-fill behaviour?

rig/machine_control/machine_controller.py
mundya: Yeah... Not sure why I didn't do that already.

mundya:
> Incidentally, do we know what the effect is of upping the loading throughput on flood-fill behaviour?

No... it passes the tests :wink:

mossblaser:
In the last few days I've been having trouble with larger SpiNNaker systems. Though this is likely to be entirely down to hardware, I'd prefer to postpone merging this branch in until things have settled down on the larger machine front as a precaution, if you don't mind?

mundya:
That's completely fine by me.

PR #169, opened 2015-07-06
Add boot parameters for larger systems.

mossblaser:
The commit changes the parameters used by the booted system to assign addresses
and P2P routes. In particular, they control the rate and frequency with which
probing and address announcement packets are sent by each monitor after
startup.

The parameters selected here are taken from an unreleased SpiNNaker 105-machine
ybug conf file. As well as making boot more reliable on larger systems, the
parameters selected disable the nearest-neighbour link probing feature which
reduces the level of background traffic in the system.

This obviously would be good to run against travis when the CI board is back home ;).


mundya:
LGTM, the board should be home from its Oxford wanderings tomorrow afternoon.

PR #171, opened 2015-07-08
Add work-around for BMP power-on bug

mossblaser:
When a power-on command is sent to a BMP on a board other than board 0, the
board does not return an acknowledgement. Though this is acknowledged as a bug
in the BMP firmware by ST, there are no plans to fix it since in practical
systems there are no scenarios where board 0 does not exist.

The fix for this is always sending power-on/off commands to board 0, whether
board 0 is in the mask or not.

**Bonus** While I was at it, this code also makes the `rig-power` command power-on all boards in a frame by default without breaking behaviour for single board setups.

### DIFF
--- a/docs/source/utility_apps.rst
+++ b/docs/source/utility_apps.rst
@@ -24,17 +24,17 @@ The ``rig-power`` command lets you quickly and easily power on and off
 SpiNNaker systems consisting of SpiNN-5 boards via their Board Management
 Processors (BMP).
 
-For example, to power cycle a single SpiNN-5 board::
+For example, to power cycle a SpiNN-5 board (or a 24-board frame there-of)::
 
     $ rig-power BMP_HOSTNAME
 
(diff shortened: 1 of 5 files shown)


docs/source/utility_apps.rst
```diff
@@ -24,17 +24,17 @@ The ``rig-power`` command lets you quickly and easily power on and off
 SpiNNaker systems consisting of SpiNN-5 boards via their Board Management
 Processors (BMP).
 
-For example, to power cycle a single SpiNN-5 board::
+For example, to power cycle a SpiNN-5 board (or a 24-board frame there-of)::
```
mundya: I'd prefer "thereof"

mundya:
Aside from nitpicking LGTM

LGTM

mossblaser:
Nitpick picked :) Thanks very much! I'll get this released momentarily.

PR #173, opened 2015-07-14
Add `rig-counters` utility

mossblaser:
This PR adds a `rig-counters` command which reports router counter value changes (e.g. dropped packets) and may be used to discover if packets are being dropped in your application.


mundya:
Awesome! :)
LGTM, thank you for doing this :)

mossblaser:
Super! Thanks for checking it out! I'll release when I get into work. Good luck with your talk!

PR #174, opened 2015-07-14
Add rig-boot 'num_boards' interface.

mossblaser:
You can use this to boot a standard arrangement of `num_boards` SpiNN-5 boards
without having to work out what dimensionality the resulting network will have.

For example:

	$ rig-boot HOSTNAME 120

Will boot a 120-board SpiNNaker system. All previous argument formats are still
supported. For example, the old-school version of the above would be:

	$ rig-boot HOSTNAME 96 60


mundya:
Looks good to me

PR #176, opened 2015-07-17
Fault tolerant rig-ps and get_machine

mossblaser:
This series of commits makes the rig-ps command and underlying get_machine() commands more robust on limping SpiNNaker machines.

A point worth repeating from 2548035 is:

> This change (strictly speaking) breaks backward compatibility as it renames
> exceptions of the RC-specific types. @mundya are you using these exceptions
> anywhere and is breaking compatibility here a problem?

Also, this commit changes the set of commands considered a "timeout" [following a discussion on c151b26](https://github.com/project-rig/rig/commit/c151b264a2aaadc41268a7891d4ed816e50a0589#commitcomment-12202415). In summary: the `RC_P2P_*` errors, with the exception of `RC_P2P_BUSY`, should be considered fatal errors, not retryable. When working on windowing to non 0, 0 chips with SARK 133 treating these as a timeout was a workaround to allow some outstanding packets to be used. Since SARK 140 should no longer require this work-around and since outstanding packets are not used with SARK 133, this should not be a problem.

### DIFF
--- a/rig/machine_control/packets.py
+++ b/rig/machine_control/packets.py
@@ -144,8 +144,9 @@ class SCPPacket(SDPPacket):
         return scp_header + self.data
 
     def __repr__(self):
-        """Produce a human-redaable summary of (the most important parts of)
-        the packet."""
+        """Produce a human-readable summary of (the most important parts of)
+        the packet.
+        """
         return ("<{} x: {}, y: {}, cpu: {}, "
                 "cmd_rc: {}, arg1: {}, arg2: {}, arg3: {}, "
                 "data: {}>".format(self.__class__.__name__,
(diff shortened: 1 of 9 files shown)


rig/machine_control/packets.py
```diff
@@ -143,6 +143,17 @@ def packed_data(self):
         # Return the SCP header and the rest of the data
         return scp_header + self.data
 
+    def __repr__(self):
+        """Produce a human-redaable summary of (the most important parts of)
```
mundya: sp.

rig/machine_control/packets.py
```diff
@@ -143,6 +143,17 @@ def packed_data(self):
...
+        the packet."""
```
mundya: Also the `""""` should be on the next lune :/

mundya:
Aside from being picky this LGTM

PR #177, opened 2015-07-18
High throughput I/O (Multiple Ethernet Connections)

### LINKED ISSUE #142: Multiple connections for boards with multiple ethernet connected chips
Not immediately required (it would be nice to have for Canada, possibly) but definitely necessary for Spaun.

As far as I can tell, the IP address of a chip can be read from the `sv` struct (`ip_addr`) and is, I assume formatted the same way as IP addresses for IP tags (e.g., `!4B`, bewaring that the default appears to be 255.255.0.0).
The real challenge is in determining the co-ordinates of likely chips, though I guess a safe approach would be to always jump 8* (north or east) from the last found ethernet chip and then read `eth_addr` to get the co-ordinates of the nearest ethernet connected chip.

I initially thought that this process should be part of `get_machine`... but now I'm not so sure.

Once we have a map of co-ordinates to ethernet addresses it becomes a drop in replacement to `MachineController.get_connection`, though we should also expose the mapping so that other software (like Nengo) can transmit SDP to SpiNNaker efficiently.

\* or 7?

Also: "bewaring" is an awesome word.

mossblaser:
When finished, this PR fixes #142. This is a branch off the fault-tollerant-rig-ps branch and is intended for merging after #176.

* [x] Facilities will be added to discover (and use) multiple Ethernet connections
* [x] A set of read_into() methods will need to be added to allow creation of the
   read buffer in advance.
* [ ] The begin_burst function will simply set a flag.
* [ ] For each connection there will be a queue of read/write requests.
* [ ] All read/write requests should be wrapped in a lambda(?) and placed in the
   buffer. For conventional reads, a buffer should be created and then
   read_into called with that buffer.
* [ ] When not in a burst, the buffer will be emptied immediately (or a fast path
   used...).
* [ ] The end_burst function will spawn some threads and in each thread work on
   processing the queues.

### DIFF
--- a/rig/geometry.py
+++ b/rig/geometry.py
@@ -5,6 +5,8 @@ import random
 
 from math import sqrt
 
+import numpy as np
+
 
 def to_xyz(xy):
     """Convert a two-tuple (x, y) coordinate into an (x, y, 0) coordinate."""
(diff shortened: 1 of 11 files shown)


mossblaser:
Some preliminary results using threads to support many streams of communication: things fairly quickly become CPU bound...

Bandwidth when using more than one connection

(Code not yet committed/pushed due to lack of formal tests and cleaning...)

mossblaser:
...as a point of comparison, running 24 instances of rig-scp in parallel achieves something slightly over 285 MBit/s without maxing out my CPU. Either way we're off my an order of magnitude...

mossblaser:
This PR has now been superceeded by #224; closing.

PR #178, opened 2015-07-29
Add SpiNN-5 board geometry functions.

mossblaser:
Adds:

* `spinn5_eth_coords`: Generate a list of board coordinates with Ethernet
  connectivity in a SpiNNaker machine.
* `spinn5_local_eth_coord`: Get the coordinates of a chip's local ethernet
  connected chip.
* `spinn5_chip_coord`: Get the coordinates of a chip on its board.
* `spinn5_fpga_link`: Get the identity of the FPGA link which corresponds with
  the supplied link.

Some of this code is taken from the numerous defunct branches aiming to improve Ethernet performance by using multiple links.

### DIFF
--- a/rig/geometry.py
+++ b/rig/geometry.py
@@ -5,6 +5,10 @@ import random
 
 from math import sqrt
 
+import numpy as np
+
+from rig.machine import Links
+
 
 def to_xyz(xy):
     """Convert a two-tuple (x, y) coordinate into an (x, y, 0) coordinate."""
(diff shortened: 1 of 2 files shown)


rig/geometry.py
```diff
@@ -214,3 +218,201 @@ def standard_system_dimensions(num_boards):
...
+        [(-4, +4), (-4, +4), (-4, +4), (-4, +4), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+8, +4), (+8, +4)],  # noqa  9
+        [(-4, +4), (-4, +4), (-4, +4), (-4, +4), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+8, +4)],  # noqa 10
+        [(-4, +4), (-4, +4), (-4, +4), (-4, +4), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8), (+4, +8)]   # noqa 11
+    ])
+], dtype=int)
+"""SpiNN-5 ethernet connected chip lookup.
+
+Used by :py:func:`.spinn5_local_eth_coord`. Given an x and y chip position
+modulo 12, return the offset of the board's bottom-left chip from the chip's
+position.
```
mundya: Just so I understand: (4, 3) -> (0, 0)?

mundya:
LGTM.

rig/geometry.py
mossblaser: So that lookup would return (-4, -3) for the key (4, 3) since the lookup gives the *offset* of the ethernet connected chip from the current chip. In this case it would indeed point at (0, 0) as the Ethernet connected chip.

I spent *so* long trying to find a closed-form version of this function and falling back on a LUT was very much a last resort (although the one used by ST and Luis).

mundya: Cool, thanks :)

PR #182, opened 2015-09-10
Use flood-fill core select packets

mundya:
Changes to SCAMP to speed up flood-fill of applications have introduced
a new packet type (flood-fill core select / FFCS) which allows several
regions to be flood-filled at the same time.  This commit enables use of
this packet type.


mundya:
Note: this switches to SCAMP 1.34.

rig/machine_control/boot.py
```diff
@@ -116,7 +116,7 @@ def boot(hostname, width, height, boot_port=consts.BOOT_PORT,
     """
     # Get the boot data if not specified.
     if boot_data is None:  # pragma: no branch
-        boot_data = pkg_resources.resource_string("rig", "boot/scamp.boot")
+        boot_data = pkg_resources.resource_string("rig", "boot/scamp-3.boot")
```
mossblaser: `scamp-134`?

mossblaser:
May be worth mentioning somewhere in a commit message that this uses features introduced in a branch in https://bitbucket.org/mundya/scamp --- even if most people don't have access, at least it will make it obvious who to ask!

rig/machine_control/machine_controller.py
```diff
@@ -731,16 +731,22 @@ def _get_next_nn_id(self):
...
         self._send_scp(
             0, 0, 0, SCPCommands.nearest_neighbour_packet,
             (NNCommands.flood_fill_start << 24) | (pid << 16) |
-            (n_blocks << 8),
-            region, sfr
+            (n_blocks << 8), 0x0, sfr
         )
 
+    def _send_ffcs(self, region, core_mask, fr):
+        """Send a flood-fill core select packet."""
```
mossblaser: Might be good to say in brief:
* This packet type was added in a branch of scamp-134
* What the packet does (as ST is the only docs and he didn't invent this, he may forget...)

rig/machine_control/machine_controller.py
```diff
@@ -764,10 +770,10 @@ def _send_ffd(self, pid, aplx_data, address):
             address += data_size
             pos += data_size
 
-    def _send_ffe(self, pid, app_id, app_flags, cores, fr):
+    def _send_ffe(self, pid, app_id, app_flags, fr):
         """Send a flood-fill end packet."""
         arg1 = (NNCommands.flood_fill_end << 24) | pid
-        arg2 = (app_id << 24) | (app_flags << 18) | (cores & 0x3ffff)
+        arg2 = (app_id << 24) | (app_flags << 18)
```
mossblaser: Worth a comment highlighting that the core set is specified by `_send_ffcs` (since looking at the scamp code may not make this obvious)?

mossblaser:
Aside from the above nitpicks about leaving a trail of breadcrumbs for future SCAMP explorers, this LGTM.

I also can't see any possible sign that the bug making only the first four FFCS packets get through is in the Python end of this...

mundya:
> I also can't see any possible sign that the bug making only the first four FFCS packets get through is in the Python end of this...

It's not only the first four, I can't find a pattern yet.

mossblaser:
Pending tests, LGTM!

mundya:
Squashing...

mundya:
>>I also can't see any possible sign that the bug making only the first four FFCS packets get through is in the Python end of this...
> 
> It's not only the first four, I can't find a pattern yet.

As a note for posterity, this problem was fixed in the version of SC&MP bundled with the final version of this PR.

PR #183, opened 2015-09-15
SameChipConstraint

mossblaser:
This new constraint allows applications to specify that a given collection of
vertices should always be placed on the same chip. This functionality is being
added to address an imminent need for this functionality in nengo_spinnaker and
a perhaps less imminent need in pinus_rigida.

@mundya and @neworderofjamie please could you confirm this will meet your future needs!

### DIFF
--- a/docs/source/place_and_route.rst
+++ b/docs/source/place_and_route.rst
@@ -83,12 +83,22 @@ Constraint
     Vertices which consume no cores are typically only useful when describing
     external devices connected to the SpiNNaker system.
     
-    Vertices which consume more than one core are typically only useful when a
-    vertex represents a group of applications which share memory. This is
-    because vertices will always be placed on a single SpiNNaker chip: they
-    cannot be split across many chips. If an application requires this type of
-    behaviour, users must perform this step in an application-defined process
-    prior to placement.
+    Vertices which consume more than one core are unlikely to be of frequent
+    use:
+    
+    * Vertices will always be placed on a single SpiNNaker chip: they cannot
+      be split across many chips. If an application requires this type of
+      behaviour, users must perform this step in an application-defined process
+      prior to placement.
+    
+    * If in an application where several cores share some on-chip resource
+      (e.g. SDRAM) and must be placed on the same chip, a
+      :py:class:`~rig.place_and_route.constraints.SameChipConstraint` can be
+      used to enforce this requirement. This has the principle advantage of
+      allowing each vertex to be connected independently via nets to other
+      vertices in the application. If a single vertex with multiple cores
+      assigned was used, all nets would be routed to and from every core
+      allocated to the vertex.
 
 :py:func:`~rig.place_and_route.wrapper`: common case wrapper
 ------------------------------------------------------------
--- a/docs/source/place_and_route.rst
+++ b/docs/source/place_and_route.rst
@@ -83,22 +83,20 @@ Constraint
     Vertices which consume no cores are typically only useful when describing
     external devices connected to the SpiNNaker system.
     
-    Vertices which consume more than one core are unlikely to be of frequent
-    use:
+    Vertices which consume more than one core are unlikely to be used
+    frequently:
     
     * Vertices will always be placed on a single SpiNNaker chip: they cannot
       be split across many chips. If an application requires this type of
       behaviour, users must perform this step in an application-defined process
       prior to placement.
     
-    * If in an application where several cores share some on-chip resource
+    * If several cores' applications must share some on-chip resource
       (e.g. SDRAM) and must be placed on the same chip, a
       :py:class:`~rig.place_and_route.constraints.SameChipConstraint` can be
-      used to enforce this requirement. This has the principle advantage of
-      allowing each vertex to be connected independently via nets to other
-      vertices in the application. If a single vertex with multiple cores
-      assigned was used, all nets would be routed to and from every core
-      allocated to the vertex.
+      used to enforce this requirement. Unlike a vertex with multiple cores,
+      each individual vertex (core) can have independent routes directly to and
+      from them.
 
 :py:func:`~rig.place_and_route.wrapper`: common case wrapper
 ------------------------------------------------------------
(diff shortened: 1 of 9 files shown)


docs/source/place_and_route.rst
```diff
@@ -83,12 +83,22 @@ Constraint
...
     external devices connected to the SpiNNaker system.
     
-    Vertices which consume more than one core are typically only useful when a
-    vertex represents a group of applications which share memory. This is
-    because vertices will always be placed on a single SpiNNaker chip: they
-    cannot be split across many chips. If an application requires this type of
-    behaviour, users must perform this step in an application-defined process
-    prior to placement.
+    Vertices which consume more than one core are unlikely to be of frequent
+    use:
```
mundya: used frequently?

docs/source/place_and_route.rst
```diff
@@ -83,12 +83,22 @@ Constraint
...
+    
+    * Vertices will always be placed on a single SpiNNaker chip: they cannot
+      be split across many chips. If an application requires this type of
+      behaviour, users must perform this step in an application-defined process
+      prior to placement.
+    
+    * If in an application where several cores share some on-chip resource
```
mundya: I don't think the "where" is necessary

docs/source/place_and_route.rst
```diff
@@ -83,12 +83,22 @@ Constraint
...
+      (e.g. SDRAM) and must be placed on the same chip, a
+      :py:class:`~rig.place_and_route.constraints.SameChipConstraint` can be
+      used to enforce this requirement. This has the principle advantage of
+      allowing each vertex to be connected independently via nets to other
+      vertices in the application. If a single vertex with multiple cores
+      assigned was used, all nets would be routed to and from every core
+      allocated to the vertex.
```
mundya: I think this could be reworded to be cleaner... but I can't write today

rig/place_and_route/place/utils.py
```diff
@@ -86,3 +91,151 @@ def apply_reserve_resource_constraint(machine, constraint):
...
+
+    for same_chip_constraint in constraints:
+        if not isinstance(same_chip_constraint, SameChipConstraint):
+            continue
+
+        # Skip constraints which don't actually merge anything...
+        if len(same_chip_constraint.vertices) <= 1:
+            continue
+
+        # The new (merged) vertex to replace the constrained vertices with
```
mundya: "The new (merged) vertex with which to replace the constrained vertices"?

"This is the type of arrant pedantry up with which I will not put"
:stuck_out_tongue:

rig/place_and_route/place/utils.py
```diff
@@ -86,3 +91,151 @@ def apply_reserve_resource_constraint(machine, constraint):
...
+        # lookup purposes)
+        merged_vertices = set(same_chip_constraint.vertices)
+
+        # Update any nets which pointed to a merged vertex
+        for net_num, net in enumerate(nets):
+            net_changed = False
+
+            # Change net sources
+            if net.source in merged_vertices:
+                if not net_changed:
```
mundya: `net_changed` can't be `True` here can it?

rig/place_and_route/place/utils.py
```diff
@@ -86,3 +91,151 @@ def apply_reserve_resource_constraint(machine, constraint):
...
+            # Change net sinks
+            for sink_num, sink in enumerate(net.sinks):
+                if sink in merged_vertices:
+                    if not net_changed:
+                        net = Net(net.source, net.sinks, net.weight)
+                    net_changed = True
+                    net.sinks[sink_num] = merged_vertex
+
+            if net_changed:
+                nets[net_num] = net
```
mundya: Given nets is already a copy there's surely no reason to modify it in place? I'd find the following clearer:

```python
def update_net(net):
    source = net.source
    if source in merged_vertices:
        source = merged_vertex

    sinks = [sink for sink in net.sinks if sink not in merged_vertices]
    if any(sink in merged_vertices for sink in net.sinks):
        sinks.append(merged_vertex)

    return Net(source, sinks, net.weight)

nets = [update(net) for net in nets]
```

mundya: Given what you're doing even:

```python
def update_net(net):
    source = net.source
    if source in merged_vertices:
        source = merged_vertex

    sinks = [sink if sink not in merged_vertices else merged_vertex
             for sink in net.sinks]

    return Net(source, sinks, net.weight)

nets = [update(net) for net in nets]
```

mundya:
This should be good for nengo_spinnaker, thank you :)

rig/place_and_route/place/utils.py
mossblaser: This is correct... I think I probably realised this when I wrote it and kept it on some odd basis but I definitely removed it along with implementing some of your other suggestions... :P

rig/place_and_route/place/utils.py
mossblaser: I'd prefer not to create new instances of every net object though premature optimisation etc... It may be better when copying the list of nets to perform a deep copy on the `Net` objects and then just modify everything straight in there.

mundya:
```python
def update_net(net):
    net_changed = False
    source = net.source
    if source in merged_vertices:
        net_changed = True
        source = merged_vertex

    sinks = list()
    for sink in net.sinks:
        if sink in merged_vertices:
            sinks.append(merged_vertex)
            net_changed = True
        else:
            sinks.append(sink)

    if net_changed:
        return Net(source, sinks, net.weight)
    else:
        return net

nets = [update(net) for net in nets]
```

mossblaser: Thanks for that :). I think I'm probably going to just keep it as is since I'm not sure the difference between the two is sufficiently great(?) Hope you don't mind!

mossblaser:
@mundya If you're happy and Travis is happy, merge at will :)

rig/place_and_route/place/utils.py
mundya: For some reason I'm disliking modifying the list by index as you loop through... I think that may just be me being oddly picky.

mundya:
Rigged

PR #186, opened 2015-10-02
Specific dependencies for Python < 3.4

mundya:
Ensures that `enum34` is only required for Python versions lower than 3.4

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -1,6 +1,7 @@
 import io
 import re
 from setuptools import setup, find_packages
+import sys
 
 
 def read_file(filename, **kwargs):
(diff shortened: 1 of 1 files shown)


setup.py
```diff
@@ -57,6 +58,12 @@ def get_new_url(url):
 with open("rig/version.py", "r") as f:
     exec(f.read())
 
+# The requirements vary depending on the version of Python
+requirements = ["numpy>1.6", "six", "sentinel", "pytz"]
+if sys.version_info[0] < 3 or sys.version_info[1] < 4:
```
mossblaser: alternatively more clearly: sys.version_info < (3, 4, 0), I think?

mundya: I didn't realise one could do that...

(Not that I'm entirely certain I know how that gets parsed)

mossblaser:
Minus the above suggestion, LGTM :)

setup.py
mossblaser: When you compare tuples it does element wise comparison from left-to-right.
Kind'a slick!

PR #187, opened 2015-10-02
Test against Python 3.5

mundya:
Adds Python 3.5 to the TravisCI test matrix.

### DIFF
--- a/DEVELOP.md
+++ b/DEVELOP.md
@@ -85,10 +85,12 @@ will also detect packaging faults (e.g. omitted binaries & support files).
 
 Rigs tests are broken up into three groups which can be run as follows:
 
-    $ py.test tests  # The Rig test-suite
-    $ py.test rig    # Doctests in rig source code
-    $ py.test docs   # Doctests in Sphinx documentation whose
-                     # filename ends with '_doctest.rst'.
+    $ py.test tests                                # The Rig test-suite
+    $ py.test rig --doctest-modules                # Doctests in rig source code
+    $ py.test docs --doctest-glob='*_doctest.rst'  # Doctests in Sphinx
+                                                   # documentation whose
+                                                   # filename ends with
+                                                   # '_doctest.rst'.
 
 Note: The first command only runs a subset of the full test suite which does not
 require an attached SpiNNaker board.
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,6 +3,7 @@ language: python
 python:
         - 2.7
         - 3.4
+        - 3.5
 install:
         - python setup.py install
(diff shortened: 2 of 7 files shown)


mundya:
I'm proposing we keep testing for Python 3.4 for a while, at least until there is an Ubuntu release with 3.5 in.

(It'd be much easier if people used Mint or Arch...)

mossblaser:
Quite... I really need to think about doing parallel tests against banks of readily available machines 'cos these tests are going to get slow!

mundya:
> (It'd be much easier if people used Mint or Arch...)

Although having said this, that just pushes the burden onto other people.  Half my morning was wiped out by the 3.4 to 3.5 transition killing my virtualenvs :(

mundya:
Apparently this does require #186

mossblaser:
I shall; though alas this means installing Python 3.5... :s

mossblaser:
...so on my machine since upgrading to 3.5 if:

```
addopts = --doctest-modules --doctest-glob='*_doctest.rst'
```

Is in my pytest.ini file the pytest process appears to get stuck in some horrifying CPU consuming loop. Looking into this...

mossblaser:
When you have a moment; please could you verify that this all looks good! Tox and pytest should now work locally too without getting into a spinloopofdeath.

DEVELOP.md
```diff
@@ -85,10 +85,12 @@ will also detect packaging faults (e.g. omitted binaries & support files).
 
 Rigs tests are broken up into three groups which can be run as follows:
 
-    $ py.test tests  # The Rig test-suite
-    $ py.test rig    # Doctests in rig source code
-    $ py.test docs   # Doctests in Sphinx documentation whose
-                     # filename ends with '_doctest.rst'.
+    $ py.test tests                                # The Rig test-suite
+    $ py.test rig --doctest-modules                # Doctests in rig source code
+    $ py.test docs --doctest-glob='*_doctest.rst'  # Doctests in Sphinx
```
mundya: Probably want to escape that asterisk somehow...

mundya:
Looks reasonable mod that Markdown error.

DEVELOP.md
mossblaser: This is just a bug in Github's markdown source rendering I think. Since it is in a literal code block it is [rendered correctly](https://github.com/project-rig/rig/blob/travis-py35/DEVELOP.md#running-tests) so doesn't need any escaping.

mundya: Cool, sorry.

mundya:
LGTM, merge away..

PR #188, opened 2015-10-02
Fix buffering

mundya:
Fixes:
 * Two typos in the machine controller tests
 * Non-contiguous writes that were within the potential range of a buffer didn't cause a flush to occur
   * Non-contiguous writes that started before the range of a buffer were ignored


mundya:
This should allow #187 to pass.

mossblaser:
LGTM! Sorry I didn't catch this in review :(

PR #189, opened 2015-10-05
Remove dead code from NER router.

mossblaser:
When the cycle-creating bug in NER was fixed in 5c4bfb7, a check got left over
from the original (broken) cycle detection code.

The code this commit removes checked, during route insertion, that the route
being inserted didn't cross an existing route segment. Thanks to 5c4bfb7, the
loop which generates the route to be inserted truncates it before it encounters
any route segments and thus the old check is no longer necessary.

This dead code was spotted by the branch-coverage checker which notably
coveralls does *not* know about and hence this code going unpunished for so
long.


mundya:
LGTM

mundya:
I merged before Travis said I could... mwhahahahahaha

PR #191, opened 2015-10-26
Allow SDRAM to be cleared after allocation

mundya:
Adds the `clear` argument to `sdram_alloc` and `sdram_alloc_as_filelike`
which will cause the allocated memory to be cleared before use.

Note: makes project-rig/nengo_spinnaker#83 unnecessary.

### DIFF
--- a/rig/machine_control/consts.py
+++ b/rig/machine_control/consts.py
@@ -67,6 +67,8 @@ class SCPCommands(enum.IntEnum):
     read = 2  # Read data
     write = 3  # Write data
 
+    fill = 5  # Fill a number of words with a given value
+
     link_read = 17  # Send a NN (or FPGA reg) read command
     link_write = 18  # Send a NN (or FPGA reg) write command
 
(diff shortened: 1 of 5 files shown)


mossblaser:
I certainly think this is a useful addition (one I have certainly wished existed) but I think it is wasteful (or at least misleading) to do it from the host. I'd suggest adding this feature IFF ST agrees to adding a flag to the allocator SCP command which will make the device do it (and then we can update this code when 140 comes out).

mundya:
> I think it is wasteful (or at least misleading) to do it from the host.

Agreed, how about a compromise along the lines of adding this implementation now but switching out for the on-chip clear if and when that becomes available?

mundya:
(Note: 140 will likely remove the need for us to clear all the routing tables manually, this seems to be a similar issue)

mossblaser:
Absolutely! I just want to be sure that it is a "when" not if :). Could you
petition ST for this feature's inclusion?

mundya:
From emails:
1) A flag for this will be included in 140
2) There is a FILL command we could use at the moment - I'll make this change to this PR now

mossblaser:
Superb. An issue needs opening flagged for the 140 tools but otherwise this LGTM and is ready to merge and release!

Issue #153, opened 2015-06-01
Use of N:M (multi-source, multi-sink) nets (possibly a severe issue)


mossblaser:
The problem
==========

After thinking a little following @mundya's #143 which makes the routing table generator merge routes which share keys and masks. I'm presuming that this above change was made in an effort to support N:M (multi-source, multi-sink) nets within Rig's net model, rather supporting some black-magic post-processing stuff and so I'll just consider N:M nets here.

Rig's "Net" model of multicast routes is that they are 1:M (single-source, multiple sink). Further the assumption is made that keys/masks are globally unique, one per net, and completely unambiguous in all situations (which use of BitFields can help ensure). This model has the nice property that it can straight-forwardly map to valid routing tables in the *general case*. (As an aside: this set of restrictions isn't clear in the docs and needs emphasising).

N:M nets can clearly be a useful thing to have: consider a case one population of neurons, `a`, is connected all-to-all with a second population, `b`. Here an N:M net could be used whose N sources were the processors simulating `a` and M sinks are the processors simulating `b`.

Clearly, such a net *can* exist in SpiNNaker, for example here's an example where the two populations are split across two processors each and reside on separate chips:

![nm net](https://cloud.githubusercontent.com/assets/193514/7914712/a6afd178-0870-11e5-8f64-baf0742d49df.png)

Indeed, this route could be generated by using two 1:M nets with the same key (after #143, anyway). There are plenty of others examples you could concoct where this approach also works correctly.

Unfortunately, N:M nets with one key/mask are *not* possible in the general case on SpiNNaker. Consider the following concrete example:

![nm broken](https://cloud.githubusercontent.com/assets/193514/7914817/45e6b2e8-0871-11e5-8ef9-576dcccffec3.png)

Here the two populations `a` and `b` are split into two processors each but this time there's an `a` and `b` processor on each chip. On the top chip, a route going South and to the `b` core would be created. On the bottom chip, a route going North and to the core with `b` would be created. The result is that any packet in the net will be infinitely bounced back-and-forth between the two chips creating duplicate packets for the two `b` cores and generally killing the network.

What next
=========

This is by no means an unknown problem and, as far as I'm aware, something all existing software has had to solve independently, by hand. Unfortunately it is cripplingly easy to fall silently into this trap accidentally with Rig, especially following #143. This duplication of effort and also the ease-of-silent-failure in Rig seems like a Bad-Thing (TM). The following solutions come to mind:

[...]

PR #192, opened 2015-10-27
Disallow repeated keys and masks

mundya:
`build_routing_tables` raises an exception if multiple nets with the exact same key and mask would cause packets to become duplicated at any node; this is generally an indication that the underlying net should be being treated as a multisource net.

This is option (3) from #153.

Clearly this is a somewhat contentious change, for example it will break `nengo_spinnaker` as stands.

### DIFF
--- a/rig/place_and_route/utils.py
+++ b/rig/place_and_route/utils.py
@@ -52,13 +52,22 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
 
     This command produces routing tables with entries optionally omitted when
     the route does not change direction (i.e. when default routing can be
-    used). Entries with identical keys and masks will be merged.
+    used).
 
-    Note: The routing trees provided are assumed to be correct and continuous
-    (not missing any hops). If this is not the case, the output is undefined.
+    .. warning::
+        A :py:exception:`~rig.place_and_route.utils.MultisourceRouteError` will
+        be raised if entries with identical keys and masks but with differing
+        routes are generated. This is not a perfect test, entries which would
+        otherwise collide are not spotted.
 
-    Note: If a routing tree has a teriminating vertex whose route is set to
-    None, that vertex is ignored.
+    .. warning::
+        The routing trees provided are assumed to be correct and continuous
+        (not missing any hops). If this is not the case, the output is
+        undefined.
+
+    .. note::
+        If a routing tree has a terminating vertex whose route is set to None,
+        that vertex is ignored.
 
     Parameters
     ----------
@@ -111,9 +120,14 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
 
             # Add a routing entry when the direction changes
             if (key, mask) in route_sets[(x, y)]:
-                # Update the existing route set if possible
+                # If there is an existing route set raise an error if the out
+                # directions are not equivalent.
+                if route_sets[(x, y)][(key, mask)].outs != out_directions:
+                    raise MultisourceRouteError(key, mask, (x, y))
+
+                # Otherwise, add the input directions as this represents a
+                # merge of the routes.
                 route_sets[(x, y)][(key, mask)].ins.add(direction)
-                route_sets[(x, y)][(key, mask)].outs.update(out_directions)
             else:
                 # Otherwise create a new route set
                 route_sets[(x, y)][(key, mask)] = _InOutPair(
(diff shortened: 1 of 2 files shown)


tests/place_and_route/test_utils.py
```diff
@@ -197,70 +199,116 @@ def test_build_routing_tables():
...
+            (Routes.west, r_west),
+            (Routes.south, RoutingTree((1, 0), {(Routes.core(1), sink1)})),
+        }),
+    }
+
+    # Create the keys
+    net_keys = {net: (0x0, 0xf) for net in (net0, net1)}
+
+    # Build the routing tables
+    with pytest.raises(Exception) as err:
```
mossblaser: Can't you assert the correct exception type here?

rig/place_and_route/utils.py
```diff
@@ -136,3 +141,18 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+    to become duplicated.
+    """
+    def __init__(self, key, mask, coordinate):
+        self.key = key
+        self.mask = mask
+        self.x, self.y = coordinate
+
+    def __str__(self):
+        return ("Two different nets with the same key ({0.key:#010x}) and "
+                "mask ({0.mask:#010x}) fork differently at ({0.x}, {0.y})".
```
mossblaser: Neat! The .format ugly syntax actually gets you something worth having! Didn't know you could do that sort of thing!

mossblaser: By the way, it may be worth ending the exception with "This is almost never what you want" or something. Maybe this check should be an optional feature (e.g. to allow redundant entries to be inserted which will later get "unmasked")?

mossblaser:
This LGTM, aside from the exception message and type check in the tests!

As mentioned offline and in #153, the issue of N:M nets is going to need to be considered properly in the near future.

rig/place_and_route/utils.py
mundya: I'm not sure I understand what you mean by "unmasked", sorry.

tests/place_and_route/test_utils.py
mundya: Yeah, you can tell I did some bad TDD here.

rig/place_and_route/utils.py
mossblaser: As in alternative redundant routes (hot spares?) could be loaded which
would be enabled by removing the other entries. Maybe?

mundya: I'm still not sure I understand; do you mean that for the last new test I added that you'd get three routing table entries at (1, 1):

(0x0, 0xf, WEST)
(0x0, 0xf, SOUTH)
(0x0, 0xf, SOUTH | WEST) -- This entry is the one that you'd get before this change

The problem then is that the table ordering is important and I believe we currently assume that tables are indifferent to ordering.

---

I can see an acceptable alternative being that the current behaviour is used if you set `strict=False` or something. If we do that I strongly believe we should at least raise a warning whenever this kind of thing occurs.

mossblaser: I'll try and explain when I'm back but yes, the strict flag would be
sufficient :)

mossblaser: Summary of offline discussion: I'm a fool and there is no real world value to allowing entries with the same key/mask pairs. No option is required here. Also the strict flag seems like an unnecessary idea.

mossblaser: ...also @mundya notes that the documentation should mention this check exists! (and that it isn't a perfect test!)

rig/place_and_route/utils.py
```diff
@@ -52,13 +52,21 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
 
     This command produces routing tables with entries optionally omitted when
     the route does not change direction (i.e. when default routing can be
-    used). Entries with identical keys and masks will be merged.
+    used).
 
-    Note: The routing trees provided are assumed to be correct and continuous
-    (not missing any hops). If this is not the case, the output is undefined.
+    .. warning::
+        An exception will be raised if entries with identical keys and masks
```
mossblaser: What type of exception? (i.e. `MultisourceRouteError`)

mundya: Fixed.

mundya:
Bump.

PR #195, opened 2015-11-03
Naive placement refactor

mossblaser:
This change extracts the kernel of the sequential hilbert placement algorithm
and makes it available as a standalone 'sequential' placer which can be
supplied with a user-defined vertex and chip use ordering. The breadth-first
vertex ordering heuristic is also broken out (and for completeness/example
usage a breadth-first sequential placement algorithm is provided).

In addition to being a generally cleaner design this will specifically make it
much easier to produce simple application-specific sequential placement
algorithms. For example when mapping a 2D grid-style application a raster-scan
sequential placement is often pretty-much what you'd want and this will make
such an algorithm very simple to implement (coming soon!).

### DIFF
--- a/rig/place_and_route/place/breadth_first.py
+++ b/rig/place_and_route/place/breadth_first.py
@@ -1,0 +1,59 @@
+"""A greedy vertex-ordering heuristic for a sequential placer."""
+
+from collections import deque, defaultdict
+
+from rig.place_and_route.place.sequential import place as sequential_place
+
+
+def breadth_first_vertex_order(vertices_resources, nets):
+    """A generator which iterates over a set of vertices in a breadth-first
+    order in terms of connectivity.
+
+    For use as a vertex ordering for the sequential placer.
+    """
+    # Special case: no vertices, just stop immediately
+    if len(vertices_resources) == 0:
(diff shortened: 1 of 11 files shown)


rig/place_and_route/place/breadth_first.py
```diff
@@ -0,0 +1,59 @@
...
+    For use as a vertex ordering for the sequential placer.
+    """
+    # Special case: no vertices, just stop immediately
+    if len(vertices_resources) == 0:
+        return
+
+    # Enumerate the set of nets attached to each vertex
+    vertex_neighbours = defaultdict(set)
+    for net in nets:
+        vertex_neighbours[net.source].update(net)
```
mossblaser: Add note: iterating over net gives the list of all vertices involved in the net.

mundya:
LGTM

rig/place_and_route/place/breadth_first.py
mundya: Other option: `vertex_neighbours[net.source].update(iter(net))` just to make it explicit?

mundya:
Cheers, all LGTM

PR #197, opened 2015-11-03
Smarter boot

mossblaser:
This PR makes two changes to the MachineController.boot() method:

* It (optionally) checks to see if the machine is already booted and does
  nothing if it is. This makes calling the method as a routine part of an
  application will no longer incur a significant runtime penalty when the
  machine is already booted. When this check is enabled it wiill also cause an
  error if you attempt to boot a BMP.
* It (optionally) checks that the machine responds to sver after it was booted.

Both optional checks are enabled by default. Though this is technically
breaking backward compatibility for anyone relying on the boot method not
sending any SCP commands, this behaviour should be sufficiently niche that this
can be considered a non-breaking change.

This commit also simplifies the rig-boot command's implementation and includes
updated documentation.

### DIFF
--- a/docs/source/control_tutorials.rst
+++ b/docs/source/control_tutorials.rst
@@ -55,6 +55,15 @@ Booting
 You can :py:meth:`~.MachineController.boot` the system like so::
 
     >>> mc.boot(12, 12)  # For a 12x12 machine
+    True
+
+If the machine could not be booted for any reason a
+:py:exc:`rig.machine_control.machine_controller.SpiNNakerBootError` will be
+raised. If no exception is raised, the machine is booted and ready to use. The
+return value of :py:meth:`~.MachineController.boot` indicates whether the
+machine was actually booted (``True``), or if it was already booted and thus
+nothing was done (``False``), most applications may consider the boot to be a
+success either way.
 
 If you're using a SpiNN-2 or SpiNN-3 board booted with no further arguments,
 only LED 0 will be usable. To enable the other LEDs, instead boot the machine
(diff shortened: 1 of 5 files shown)


docs/source/control_tutorials.rst
```diff
@@ -55,6 +55,15 @@ Booting
 You can :py:meth:`~.MachineController.boot` the system like so::
 
     >>> mc.boot(12, 12)  # For a 12x12 machine
+    True
+
+If the machine could not be booted for any reason a
+:py:exc:`rig.machine_control.machine_controller.SpiNNakerBootError` will be
+raised. If no exception is raised, the machine is booted and ready to use. The
+return value of :py:meth:`~.MachineController.boot` indicates whether the
+machine was actually booted (``True``), or if it was already booted thus
```
mundya: "and thus"?

rig/machine_control/machine_controller.py
```diff
@@ -199,27 +198,18 @@ def _send_scp(self, x, y, p, *args, **kwargs):
         connection = self._get_connection(x, y)
         return connection.send_scp(length, x, y, p, *args, **kwargs)
 
-    def boot(self, width, height, **boot_kwargs):
+    def boot(self, width, height, force=False, check=True, **boot_kwargs):
```
mundya: I'm not sure `force` is a good parameter name, to me it implies that a machine can be forced to boot.

mundya: How about `check_before_boot` and `check_boot_successful`?

mossblaser: `only_if_needed` and `check_booted`?

mundya: Yes :)

mossblaser:
Those alternative names are much nicer and much more pleasant to read! Thanks for that! Just awaiting your final approval and a merge. (It would be worth waiting for travis to pass everything first given the nature of these changes!)

PR #199, opened 2015-11-08
Fix PyPI badge

mossblaser:
Switched to `shields.io` as badge provider.

Clearly this is a critical bugfix... It has been bothering me for a good while now though...


mundya:
Truly critical. Major point bump worthy?

mossblaser:
Hah! :P Perhaps even worth forking!

PR #201, opened 2015-11-15
Discover and use multiple Ethernet connections

mossblaser:
This commit adds support for discovering and using multiple Ethernet
connections (when available) but does not feature any way to use these
connections in parallel to improve performance. A limited performance
improvement due to reduced average latency within the machine may be possible.
This feature may also allow more robust communication with machines whose links
have died during runtime.

This commit was cherry-picked from commit 08ff6e1 (part of the
high-throughput-io branch) since its functionality is useful and doesn't
involve any API changes.

Perhaps some mechanism for triggering this discovery automatically would be worthwhile since there is very little reason to *not* want this functionality(?)

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -18,6 +18,8 @@ from rig.machine_control.scp_connection import SCPConnection, SCPError
 from rig import routing_table
 from rig.machine import Cores, SDRAM, SRAM, Links, Machine
 
+from rig.geometry import spinn5_eth_coords, spinn5_local_eth_coord
+
 from rig.utils.contexts import ContextMixin, Required
 from rig.utils.docstrings import add_signature_to_docstring
 
@@ -108,10 +110,20 @@ class MachineController(ContextMixin):
                                                         "boot/sark.struct")
             self.structs = struct_file.read_struct_file(struct_data)
 
-        # Create the initial connection
-        self.connections = [
-            SCPConnection(initial_host, scp_port, n_tries, timeout)
-        ]
+        # This dictionary contains a lookup from chip (x, y) to the
+        # SCPConnection associated with that chip. The special entry with the
+        # key None is reserved for the connection initially made to the
+        # machine and is special since it is always known to exist but its
+        # actual position in the network is unknown.
+        self.connections = {
+            None: SCPConnection(initial_host, scp_port, n_tries, timeout)
+        }
+
+        # The dimensions of the system. This is set by discover_connections()
+        # and is used by _get_connection to determine which of the above
+        # connections to use.
+        self._width = None
+        self._height = None
 
     def __call__(self, **context_args):
         """For use with `with`: set default argument values.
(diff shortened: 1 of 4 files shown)


mossblaser:
...apparently GitHub decided to garble my labels...

rig/machine_control/machine_controller.py
```diff
@@ -174,7 +186,20 @@ def send_scp(self, *args, **kwargs):
...
+        if self._width is None or self._height is None:
+            return self.connections[None]
+        else:
+            # If possible, use the local Ethernet connected chip
+            eth_chip = spinn5_local_eth_coord(x, y, self._width, self._height)
+            conn = self.connections.get(eth_chip)
+            if conn is not None:
+                return conn
+            else:
+                # If no connection was available to the local board, chose
```
mundya: "choose"

rig/machine_control/machine_controller.py
```diff
@@ -174,7 +186,20 @@ def send_scp(self, *args, **kwargs):
...
+                # another arbitrarily.
+                # XXX: This choice will cause lots of contention in systems
+                # with many missing Ethernet connections.
```
mundya: ~~It wouldn't be too hard to construct something that picked ...~~

rig/machine_control/machine_controller.py
```diff
@@ -307,6 +332,63 @@ def boot(self, width, height,
...
+                ip = self.get_ip_address(x, y)
+                if ip is not None:
+                    # Create a connection to the IP
+                    self.connections[(x, y)] = \
+                        SCPConnection(ip, self.scp_port,
+                                      self.n_tries, self.timeout)
+                    # Attempt to use the connection (and remove it if it
+                    # doesn't work)
+                    try:
+                        self.get_software_version(x, y)
```
mundya: I'm wondering whether being able to change the number of retries or the timeout would be useful for this method?

mossblaser: Not being able to change the number of tries is a general pain in various places (see the new boot code for example which creates a whole new MachineController just for that!). Not sure what to do about it really either...

mundya: I'll have a look this afternoon and see what I come up with.

mossblaser: Since it should be very rare that this process should fail (ST regards an ethernet-connected-chip failure as fatal) are you happy to merge this "slow in the face of errors" version for now?

mundya: Yeah, that's fine. Merge away! I managed to get something not too unpleasant cobbled together for changing the number of retries. I'll push that as soon as I get internet again tomorrow.

mossblaser: Thanks!

rig/machine_control/machine_controller.py
```diff
@@ -307,6 +332,63 @@ def boot(self, width, height,
         return True
 
     @ContextMixin.use_contextual_arguments()
+    def discover_connections(self, x=0, y=0):
```
mundya: @mossblaser - do you have any feelings about this method being called by `get_system_info`? I realised that I've not been calling this method when it would probably be in my interests to do so.

mossblaser: I'd been wondering when/how/if to call it automatically and couldn't think of a good place. I'm not fond of something like get_system_info having such a "magic" side-effect. Any other ideas?

PR #205, opened 2015-11-24
Fix error in read/write across links due to bad merge

mossblaser:
Originally spotted by @mundya


mundya:
LGTM, merge when ready!

Issue #202, opened 2015-11-15
Use only unused cores


mossblaser:
# Introduction

~~In the beginning there was read and write and it was good. Except when multiple things wanted to share memory. Then it was very, very bad. And so on the 1.33rd day, S-t created SDRAM alloc and life was good. Multiple applications and even the monitor could share the chip-wide memory resources, there was much rejoicing about entering the 21st century and all was well.~~

~~...but all was not well in the world. All this time ReserveResourceConstraints were all that kept Rig applications from trampling the monitor; and nothing at all protected the poor, innocent applications already running on the machine!~~

Ahem... Anyway... It seems odd that we're so careful to allocate SDRAM in some sane way and yet do not take the same approach to application cores. As well as being a little unsightly on a purely perfectionist perspective, here are a few examples of where this behaviour makes life hugely unpleasant:

* If the on-chip monitor process ever grows to multiple cores, every Rig *application* will need patching -- yuck! (This is a real possibility that ST/Luis are considering for the future.)
* If you want packet reinjection you *have* to integrate it into your application. In a sane world you'd start the packet reinjector on every core (using ybug, some standalone script, or whatever) and then just run your application as normal. Doing that right now will result in an unreported massacre of random instances of the packet-reinjector.
* If you want to do some detailed logging of network resources (e.g. to spot burstiness) the current best solution is to hack your application to reserve core 1 and then hope your logging tool gets loaded onto core 1... If anything goes wrong the result is a diagnostics-free mess.
* Insert many other instrumentation/debug-like scenarios here...
* Seriously, why does every application developer need to specify -- *by hand* -- that core 0 should not be used?!

Either way, the current way of doing things is a bit short-sighted and I think fixing this warrants proper consideration. This issue describes a suggested fix for this.

# Proposed fix

I've been thinking about this for a few days now and so far the nicest solution to this seems to be to introduce a `get_resource_constraints()` method to MachineController (maybe needs better name...) which produces a whole load of ReserveResourceConstraints which rope-off any cores which aren't idle. ST would be amenable to adding a bit-mask of idle cores in SC&MP to make this process more efficient than reading the process table.

While we're at it we should really try and do something similar for SDRAM availability (e.g. if an instrumentation app has used 32 MByte of SDRAM per chip you probably can't afford to ignore this when placing memory-bound applications). Obviously this will require some changes to SC&MP, most likely with assistance or at least guidance from ST.

## Pros of this method

[...]

PR #206, opened 2015-11-26
Only use idle cores

mossblaser:
This PR began as an answer to #202 to attempt to automatically avoid using non-idle cores and has resulted in a few wider-scoped changes to make this process cleaner.

* `get_machine` has been deprecated since having the MachineController directly handling P&R artefacts seems like a bad idea in general.
* `get_chip_info` uses a new SCP command to quickly return basic information about the health and state of a chip.
* A new method, `get_system_info`, and associated `SystemInfo` container has been added which uses `get_chip_info` to quickly gather the state of every chip, link and core in a SpiNNaker machine. The `SystemInfo` contains a superset of the information `Machine` did and does so in a far more explicit way. This call is also a little quicker than `get_machine` was.
* A utility function `build_machine` has been added which builds a `Machine` object from a `SystemInfo` when you are doing place-and-route work.
* A utility function `build_core_constraints` has been added which builds P&R constraints which reserve all non-idle cores in a machine (including monitors) using information in a `SystemInfo` object.
* The old place-and-route wrapper has been deprecated in favour of a new wrapper which doesn't place things on non-idle cores. It also takes a `SystemInfo` object in place of a `Machine`.
* Machine and its associated sentinels now live in `rig.place_and_route.machine` and are aliased in `rig.place_and_route`.
* Links now lives in `rig.links`.
* For backward compatibility, wrappers exist for the old functions, methods and API structures, the use of which will trigger a deprecation warning.

Once this PR is merged and released, I'll submit patches for other projects.
  * Nengo SpiNNaker
  * pinus_rigida
  * network_tester
  * rig_examples


rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
             Links are reported as dead when the device at the other end of the
-            link does not respond to SpiNNaker nearest neighbour packets. This
-            may thus mistakenly report links attached to peripherals as dead.
+            link is not a SpiNNaker chip. Thus links attached to peripherals
+            are always marked as dead.
 
         .. note::
-            The probing process does not report how much memory is free, nor
-            how many processors are idle but rather the total available.
+            The returned object not report how much memory is free, nor how
```
mundya: "object does *not* report"

mossblaser:
OK Since this PR is fully backward compatible with older versions of Rig I'd suggest merging it in and I'll work on PRs for other projects in due-course since this PR won't break anything.

rig/place_and_route/wrapper.py
```diff
@@ -1,15 +1,131 @@
...
+    application_map = build_application_map(vertices_applications, placements,
+                                            allocations, core_resource)
+
+    # Build data-structures ready to feed to the machine loading functions
+    routing_tables = build_routing_tables(routes, net_keys)
+
+    return placements, allocations, application_map, routing_tables
 
 
 def wrapper(vertices_resources, vertices_applications,
```
mundya: https://docs.python.org/3/library/exceptions.html#DeprecationWarning ?

mundya: Damn, you were way ahead of me, sorry. I thought there was a decorator for this somewhere!

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
+    @ContextMixin.use_contextual_arguments()
     def get_working_links(self, x, y):
         """Return the set of links reported as working.
 
-        The returned set lists only links over-which nearest neighbour
-        peek/poke commands could be sent. This means that links connected to
-        peripherals may falsely be omitted.
+        This command tests each of the links leaving a chip by sending a PEEK
+        nearest-neighbour packet down each link to verify that the remote
+        device is a SpiNNaker chip. If no reply is receieved via a given link
```
mundya: sp. received

rig/place_and_route/wrapper.py
mossblaser: Very irritatingly there is not (though of course there are 100s of slightly different and often broken ones floating around the web). Additionally, warnings do not show up without `-Wa` on your Python interpreter... Sigh... Ah well! It is probably good enough given the entire userbase of Rig has been following this saga in person...

docs/source/place_and_route.rst
```diff
@@ -98,14 +98,21 @@ Constraint
...
-do exactly this with a minimum of fuss. For more advanced users, this function
-can be used as an example of the complete place-and-route process.
+routing tables. For most users the
+:py:func:`rig.place_and_route.place_and_route_wrapper` will do exactly this
+with a minimum of fuss. For more advanced users, this function can be used as
+an example of the complete place-and-route process.
+
+.. autofunction:: rig.place_and_route.place_and_route_wrapper
+
+The above function replaces the very similar but now deprecated
```
mundya: `.. note::`?

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
+        # Calculate the extent of the system
+        max_x = max(x_ for (x_, y_), r in iteritems(p2p_tables)
+                    if r != consts.P2PTableEntry.none)
+        max_y = max(y_ for (x_, y_), r in iteritems(p2p_tables)
+                    if r != consts.P2PTableEntry.none)
+
+        chips_info = {}
+
+        for (x, y), p2p_route in iteritems(p2p_tables):
+            if x <= max_x and y <= max_y:
```
mundya: I don't think I understand why this test is necessary given that you also check `p2p_route != ....none` below

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
+    def get_machine(self, x=0, y=0, default_num_cores=18,
+                    _all_chips_info=None):
         """Probe the machine to discover which cores and links are working.
 
         .. note::
             Links are reported as dead when the device at the other end of the
-            link does not respond to SpiNNaker nearest neighbour packets. This
-            may thus mistakenly report links attached to peripherals as dead.
+            link is not a SpiNNaker chip. Thus links attached to peripherals
+            are always marked as dead.
```
mundya: I feel this needs clarifying to the extent that a link is reported as dead *when it is dead* or when the device at the other end is not a SpiNNaker chip. But I could just be being really picky, it is my way.

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
 
         .. note::
-            The probing process does not report how much memory is free, nor
-            how many processors are idle but rather the total available.
+            The returned object does not report how much memory is free, nor
+            how many cores are idle but rather the total available. See
```
mundya: Should this not document that it reports the largest free block of memory available, reporting 32MB where there's actually a lot of fragmented 16MB blocks around is fine if I only ever need memory in 16MB blocks. Of course, this is a hard requirement to express.

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1659,120 @@ def get_p2p_routing_table(self, x, y):
...
+            The chip (x, y) supplied is the one where the search for working
+            chips begins. Selecting anything other than (0, 0), the default,
+            may be useful when debugging very broken machines.
 
         Parameters
         ----------
         default_num_cores : int
             The number of cores generally available on a SpiNNaker chip
             (including the monitor).
+        _all_chips_info : None
```
mundya: I think this should be under `Other Parameters`, and should be documented as `None or ...`

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
-                        for link in set(Links) - working_links:
-                            dead_links.add((x, y, link))
-                    except SCPError:
-                        # The chip was listed in the P2P table but is not
-                        # responding. Assume it is dead anyway.
-                        dead_chips.add((x, y))
+                    if chip_info.num_cores < default_num_cores:
+                        resource_exception = chip_resources.copy()
+                        resource_exception[Cores] = min(default_num_cores,
+                                                        chip_info.num_cores)
```
mundya: But you already know that `chip_info.num_cores` < `default_num_cores`

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
                        chip_resource_exceptions,
                        dead_chips, dead_links)
 
+    @ContextMixin.use_contextual_arguments()
+    def get_resource_constraints(self, machine,
+                                 core_resource=Cores,
+                                 sdram_resource=SDRAM,
+                                 sram_resource=SRAM,
+                                 _all_chips_info=None):
+        """Returns a set of place-and-route
```
mundya: As per PEP257:
"The docstring is a phrase ending in a period. It prescribes the function or method's effect as a command ("Do this", "Return that"), not as a description; e.g. don't write "Returns the pathname ...". "

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+
+        This method, like :py:meth:`get_machine`, is aimed at users of Rig's
+        place-and-route tools and assumes that one resource type represents
+        Cores while another represents SDRAM.
+
+        The returned set of
+        :py:class:`~rig.place_and_route.constraints.ReserveResourceConstraint`s
+        reserves all cores not in an Idle state (i.e. not a monitor and not
+        already running an application).
+
```
mundya: There probably shouldn't be a paragraph break here (given the starting 2 words of the next sentence).

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        This set also reserves ranges of SDRAM/SRAM starting at the top
```
mundya: This set also contains reservations to SDRAM/SRAM to account for any applications already making use of the memory.?

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        (highest address) to reduce the resource availability on each chip to
+        approximately the amount actually available in the machine, accounting
+        for any already-running applications. Note that due to memory
+        fragmentation this constraint will make a conservative estimate based
+        on the largest free block of SDRAM/SRAM in the heap.
```
mundya: To account for memory fragmentation the reservation made will be a conservative estimate based upon the size of the largest free block of SDRAM/SRAM.

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+
+        Parameters
+        ----------
+        machine : :py:class:`~rig.machine.Machine`
+            The :py:class:`~rig.machine.Machine` model of the current SpiNNaker
+            machine (e.g. from :py:meth:`get_machine`).
+        core_resource : resource (Default: :py:data:`~rig.machine.Cores`)
+            The resource identifier used for cores. If using
+            :py:meth:`get_machine`, the default correct.
```
mundya: "the default correct"?

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        sdram_resource : resource (Default: :py:data:`~rig.machine.SDRAM`)
+            The resource identifier used for SDRAM. If using
+            :py:meth:`get_machine`, the default correct.
```
mundya: likewise

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        sram_resource : resource (Default: :py:data:`~rig.machine.SRAM`)
+            The resource identifier used for SRAM. If using
+            :py:meth:`get_machine`, the default correct.
```
mundya: likewise

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        _all_chips_info : None
```
mundya: Again, I think this should be documented more fully in `Other Parameters`

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        :py:class:`rig.place_and_route.constraints.ReserveResourceConstraint`,\
+        ...]
+            A set of place-and-route constraints which reserves non-idle cores
+            and SDRAM. The resource types given in the ``core_resource`` and
+            ``sdram_resource`` arguments will be reserved accordingly.
+        """
+        if _all_chips_info is not None:
+            chips_info = _all_chips_info[2]
+        else:
+            chips_info = {}
```
mundya: `collections.defaultdict(lambda (x, y): self.get_chip_info(x, y))`?

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+                            (x, y))
+                    else:
+                        cur_reservation.reservation = slice(
+                            cur_reservation.reservation.start,
+                            core + 1)
+            if cur_reservation is not None:
+                constraints.append(cur_reservation)
+
+            # Reserve enough SDRAM such that the amount remaining will be the
+            # size of the largest free block in the heap.
```
mundya: We're relying on there being little fragmentation. Is that wise?

[NB I think it is probably fine, assuming that SARK behaves sanely when freeing and that most applications can be expected to make few, large, requests for memory]

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+                (x, y)))
+
+            # Reserve enough SRAM such that the amount remaining will be the
+            # size of the largest free block in the heap.
+            total_sram = machine[(x, y)][sram_resource]
+            avail_sram = chip_info.largest_free_sram_block
+            constraints.append(ReserveResourceConstraint(
+                sram_resource,
+                slice(avail_sram, total_sram),
+                (x, y)))
```
mundya:
```python
def make_memory_reservation(resource, available):
    total = machine[(x, y)][resource]
    return ReserveResourceConstraint(resource, slice(available, total), (x, y))

constraints.append(make_memory_reservation(sdram_resource, chip_info.largest_free_sdram_block))
constraints.append(make_memory_reservation(sram_resource, chip_info.largest_free_sram_block))
```

To reduce some duplication?

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+
+        return constraints
+
+    def get_machine_and_resource_constraints(self, x=0, y=0,
+                                             default_num_cores=18):
+        """An efficient wrapper around both :py:meth:`.get_machine` and
+        :py:meth:`.get_resource_constraints`.
```
mundya: "Get a representation of the machine and constraints upon the resources which may be used."... or similar

rig/machine_control/machine_controller.py
```diff
@@ -1739,34 +1807,169 @@ def get_machine(self, x=0, y=0, default_num_cores=18):
...
+        Returns
+        -------
+        (machine, resource_constraints)
+            machine is a :py:class:`rig.machine.Machine` describing what
+            resources exist on each chip (see :py:meth:`.get_machine`).
+
+            resource_constraints is a list of place-and-route
+            :py:class:`~rig.place_and_route.constraints` which restrict the
+            resources described in the machine to only those which are not
+            already in use.
```
mundya:
```
Returns
------
:py:class:`rig.machine.Machine`
    Description of the resources that exist on each chip.
[:py:class:`~rig.place_and_route.constraints`]
    Restrictions to the resources described in the machine to account for current usage of those resources.

See Also
--------
 * :py:meth:`.get_machine`
 * :py:meth:`.get_resource_constraints`

File: rig/place_and_route/wrapper.py [hunk 23]
```diff
@@ -1,15 +1,131 @@
...
+def place_and_route_wrapper(vertices_resources, vertices_applications,
+                            nets, net_keys,
+                            machine, constraints=[],
+                            place=default_place, place_kwargs={},
+                            allocate=default_allocate, allocate_kwargs={},
+                            route=default_route, route_kwargs={},
+                            core_resource=Cores, sdram_resource=SDRAM):
+    """Wrapper for core place-and-route tasks for the common case.
+
+    At a high level this function essentially takes a set of vertices and nets
```
mundya: ~~At a high level~~ ... ~~essentially~~

mundya:
I still need to read the tests... sorry

mossblaser:
Thanks very much for all the reviewing, take your time! I'll get on the corrections...

I've been thinking more about all this and I've come to the conclusion some things about the API coming out of this PR leave me feeling quite uncomfortable:

With the API exposed as it is here you cannot trivially answer the question "how many idle cores are there in total" without doing your own get_chip_info which is wasteful and doesn't feel right. As a result I think the `_get_all_chips_info` method should be made public. It should probably get a better name too...

I do not like the fact that `get_resource_constraints`, a MachineController method, is returning place-and-route constraints, a construct designed for a very specific application and with little general use outside that context. Additionally, if you already have a `ChipInfo` for every chip there is no reason for `get_resource_constraints` to communicate with the machine and thus no reason for it to be part of MachineController. Instead I believe this should live with the other place-and-route utility functions.

Perhaps more controversially `get_machine` is essentially exactly the same in that once given a `ChipInfo` it does not actually need to communicate with the machine at all. Though the Machine object is carefully designed to actually be quite useful outside of place-and-route, it still has plenty of quirks related to the place-and-route infrastructure (e.g. the concept of 'resources' and the Cores, SDRAM and SRAM sentinels). All this points towards `get_machine` being pulled out and put into a place-and-route utility function. Obviously a deprecation-flagged wrapper in MachineController must remain!

Sadly going from the `Machine` object to a bare dictionary of `ChipInfo`s feels like a regression in high-level functionality in many ways. To relieve some of this I'd propose making `get_all_chips_info` return a dict-like object with extra functionality beyond the usual lookup from chip to `ChipInfo`. The API should hopefully recreate the useful functions of the Machine object, namely easy iteration over (live) chip and link coordinates, easy testing for chip and link liveness, directly getting the width/height of the network. Further functionality which wouldn't have been appropriate for Machine objects would be things like iterating over core coordinates, or even cores in a particular state.

So, what do you think?

* Is the above sensible?
* If so, what should the chip-information-holding object be called? It needs to be short/snappy and *not* be confused for `Machine`...

docs/source/place_and_route.rst
mossblaser: I've actually just removed this line entirely since the docstring for `place_and_route_wrapper` actually contains this and it even is a `.. note::` ;).

rig/machine_control/machine_controller.py
mossblaser: This line is a throwback from the darker days when `get_p2p_routing_table` actually grabbed the whole table rather than just the portion for the machine which actually exists. As such it should, of course, be removed but first I need to work out why it doesn't fail the branch coverage checker...

rig/machine_control/machine_controller.py
mossblaser: You don't; it may be equal; `chip_info.num_cores` is the number working of cores present (which is usually 17 or 18), not just the number which are idle (which generally would be strictly < 18).

rig/machine_control/machine_controller.py
mossblaser: The factory function to a defaultdict doesn't get given any arguments so sadly this is not possible.

rig/machine_control/machine_controller.py
mossblaser: > We're relying on there being little fragmentation. Is that wise?

I think so.

> most applications can be expected to make few, large, requests for memory

Indeed. Further due to the fact that interleaving arbitrary applications is highly unlikely to be possible (e.g. due to routing key related issues) multi-application use will probably limited to instrumentation and debug stuff. Combined with the fact that real applications are unlikely to allocate/free resources during runtime, the memory alloc/free pattern is pretty much ideal in terms of minimising fragmentation.

> assuming that SARK behaves sanely when freeing

I believe this is the case. According to ST when a block is freed it is merged with any neighbouring free blocks if possible.

rig/machine_control/machine_controller.py
mundya: Regardless of the meaning of `x` and `y`:

```
1816:    if x < y:
1819:        z = min(x, y)
```

`z` will always equal `x`.

rig/machine_control/machine_controller.py
mundya: Sorry, I should have known that.  It's a shame...

mossblaser:
OK; following the above the following things have happened:

* A new `get_system_info` method replaces `get_machine`, `get_resource_constraints` and all that nonsense.
* A new `SystemInfo` type has been introduced which is a dict from (x, y) to ChipInfo along with some utility methods for iterating over things and checking membership. It should replace Machine objects in settings not related to P&R.
* `get_machine` is now deprecated (but still present for backward compatibility).
* A pair of P&R utility methods `build_machine` and `build_core_constraints` have been provided which use a `SystemInfo` object to build those data structures.
* The new `place_and_route_wrapper` now takes a `SystemInfo` as an argument.

I think some reorganisation still needs to happen with the Rig module hierarchy:

* [ ] The `Machine` type and associated sentinels should be moved into the `place_and_route` namespace.
* [ ] The `Links` type probably needs a better home than the only member of `rig/machine.py`...

mossblaser:
OK, I think this PR is now finally complete and ready for final review and merging! The comment at the top of the PR should hopefully enumerate the changes made. Sadly due to the light internal reorganisation this PR has touched more code than it would otherwise need to.

rig/machine_control/machine_controller.py
```diff
@@ -1657,51 +1661,153 @@ def get_p2p_routing_table(self, x, y):
...
+            A named tuple indicating the number of working cores, the states of
+            all working cores, the set of working links and the size of the
+            largest free block in SDRAM and SRAM.
+        """
+        info = self._send_scp(x, y, 0, SCPCommands.info, expected_args=3)
+        num_cores = info.arg1 & 0x1F
+        core_states = [consts.AppState(c)
+                       for c in struct.unpack("<18B", info.data)]
+        working_links = set(link for link in Links
+                            if (info.arg1 >> (8 + link)) & 1)
```
mundya: What's the performance of this like given that it's called for every chip?

rig/machine_control/machine_controller.py
```diff
@@ -1797,6 +1868,173 @@ class CoreInfo(collections.namedtuple(
...
+    'ChipInfo', "num_cores core_states working_links "
+                "largest_free_sdram_block,largest_free_sram_block")):
+    """Information returned about a chip.
+
+    Parameters
+    ----------
+    num_cores : int
+        The number of working cores on the chip.
+    core_states : [:py:class:`~rig.machine_control.consts.AppState`, ...]
+        The state of each working core in the machine in a lust ``num_cores``
```
mundya: lust -> list

rig/machine_control/machine_controller.py
mossblaser: This set-comprehension? I'm not sure but the new `get_system_info` runs faster than the old `get_machine`. I'd be very disappointed if anything but the network round-trips were dominating runtime at the moment. It would be worth investigating appropriate ways to make all those SCP commands get sent in parallel in the future...

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+            >> sys_info = mc.get_system_info()
+            >> machine = build_machine(sys_info)
+
+    Parameters
+    ----------
+    system_info : :py:class:`rig.machine_control.machine_controller.SystemInfo`
+        The resource availability information for a SpiNNaker machine,
+        typically produced by
+        :py:meth:`rig.machine_control.MachineController.get_system_info`.
+    core_resource : resource (default: :py:class:`rig.machine.Cores`)
```
mundya: "`rig.place_and_route.Cores`"?

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+        The resource type to use to represent the number of working cores on a
+        chip, including the monitor, those already in use and all idle cores.
+    sdram_resource : resource (default: :py:class:`rig.machine.SDRAM`)
```
mundya: Likewise.

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+        The resource type to use to represent SDRAM on a chip. This resource
+        will be set to the number of bytes in the largest free block in the
+        SDRAM heap. This gives a conservative estimate of the amount of free
+        SDRAM on the chip which will be an underestimate in the presence of
+        memory fragmentation.
+    sram_resource : resource (default: :py:class:`rig.machine.SRAM`)
```
mundya: Likewise.

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+        The resource type to use to represent SRAM (a.k.a. system RAM) on a
+        chip. This resource will be set to the number of bytes in the largest
+        free block in the SRAM heap. This gives a conservative estimate of the
+        amount of free SRAM on the chip which will be an underestimate in the
+        presence of memory fragmentation.
+
+    Returns
+    -------
+    :py:class:`rig.machine.Machine`
+        A :py:class:`~rig.machine.Machine` object representing the
```
mundya: Likewise.

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+    running an application).
+
+    .. note::
+
+        Historically it was required that a
+        :py:class:`~rig.place_and_route.constraints.ReserveResourceConstraint`
+        was manually added by every application which reserved the monitor core
+        on each chip. Improves on this by automatically generating appropriate
+        constraints for reserving not just monitor cores but also other cores
+        which are already in use.
```
mundya: "Historically, every application was required to add a :py:class:`~rig.place_and_route.constraints.ReserveResourceConstraint` to reserve the monitor processor on each chip. This method improves upon this approach by automatically generating constraints which reserve not just the monitor core but also any other cores which are already in use."

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+        The :py:class:`~rig.machine.Machine` model of the current SpiNNaker
+        machine (e.g. from :py:func:`.build_machine`).
+    core_resource : resource (Default: :py:data:`~rig.machine.Cores`)
+        The resource identifier used for cores.
+
+    Returns
+    -------
+    [:py:class:`rig.place_and_route.constraints.ReserveResourceConstraint`, \
+            ...]
+        A set of place-and-route constraints which reserves all non-idle. The
```
mundya: "all non-idle cores"

rig/place_and_route/utils.py
```diff
@@ -4,13 +4,187 @@
...
+        resource type given in the ``core_resource`` argument will be reserved
+        accordingly.
+    """
+    constraints = []
+
+    for x, y in machine:
+        chip_info = system_info[(x, y)]
+
+        # Construct a minimal set of ReserveResourceConstraints which
+        # reserve the cores which are not in the idle state.
```
mundya: Minimal only within chips, right? For a n-chip machine this will generate n reservations for the monitor processors won't it?

tests/machine_control/test_machine_controller.py
```diff
@@ -2171,97 +2236,114 @@ def test_get_num_working_cores(self, num_cpus):
...
+                             [((18 | (0b111110 << 8)),
+                               18,
+                               set(l for l in Links if l != Links.east)),
+                              ((18 | (0b011111 << 8)),
+                               18,
+                               set(l for l in Links if l != Links.south)),
+                              ((17 | (0b111111 << 8)),
+                               17,
+                               set(Links)),
+                              ])
```
mundya:
```python
@pytest.mark.parametrize(
    "arg1, num_cores, working_links",
...
```

Would be a *little* more readable...

rig/machine_control/machine_controller.py
mundya: > This set-comprehension?

I meant in general, but was just intrigued whether `info.arg1 & (0x100 << link)` would perform better. That said, I agree that the network *should* dominate...

> I'm not sure but the new get_system_info runs faster than the old get_machine.

Awesome :)

mundya:
My read through the tests was cursory, but this looks mostly good to me.  I am concerned that on big machines it'll generate huge number of resource constraints, but I may just be misunderstanding that bit of code.

rig/place_and_route/utils.py
mossblaser: Ugh... my regexing for these has definitely failed... :(

rig/place_and_route/utils.py
mossblaser: This is certainly true. I was debating what to do about global reservations... You'd have to iterate over the chips twice and do a bitwise-and of the non-idle cores, globally reserve that and then go through everything again reserving all the exceptions.

In terms of the runtime expense of this for the allocation/placement algorithms, generating one global constraint vs many local ones has little or no performance difference. I suppose really I should do global things properly, if nothing else but for human readability...

tests/machine_control/test_machine_controller.py
mossblaser: Sadly even doing that does not allow the stimulus lines to fit on one line so it probably isn't worth changing in this case. In general though I'd agree!

mossblaser:
> I am concerned that on big machines it'll generate huge number of resource constraints, but I may just be misunderstanding that bit of code.

See previous comment.

TODOs for me before I merge this then are:

* [ ] Fix misc things pointed out above
* [ ] Fix remaining references to "rig.machine"
* [ ] Enable generation of global constraints when possible

mundya:
>  Enable generation of global constraints when possible

Would an option be allowing constraints to be associated with multiple locations? - That way you could just modify one set of constraints and then mark the constraint as global (remove the location) if the difference between the set of locations the constraint is applied to and the overall set of locations is just the set of dead chips.

`new_location = None if (old_locations - all_locations == dead_locations) else old_locations`

mossblaser:
> Would an option be allowing constraints to be associated with multiple locations?

I don't think that would meaningfully save you compute or storage costs but would make life more complex. I'll just stick to producing a minimal set of constraints as they stand now.

mossblaser:
OK; this awaits your final review. Feel free to merge if you're happy!

mundya:
LGTM, I'll await the tests coming back

mundya:
Are the changes to `rig_par_diagram` necessitated by this trivial to make?

mossblaser:
Hopefully none what so ever. Has it started failing?

mundya:
No, I'm stupid, I'd forgotten that `Machine` still existed. Sorry!

PR #207, opened 2015-11-26
Save some memory in `RegionTree`

mundya:
Drop references to sub-trees once they're filled up.


mundya:
Ah, it works fine enough

PR #210, opened 2015-12-07
Changes to RegionTree for FFCS

mundya:
Minimising flood-fill regions is hard. This commit changes the RegionTree structure such that it:

 - Will generate region and core masks in the order required by SCAMP (that is, in increasing order of ``region << 32 | coremask``)
 - Has some knowledge of cores, and so can perform a bit more minimisation internally.

### DIFF
--- a/rig/machine_control/regions.py
+++ b/rig/machine_control/regions.py
@@ -159,7 +159,7 @@ class RegionCoreTree(object):
             # Iterate through the subregions and recurse, we iterate through in
             # the order which ensures that anything we yield is in increasing
             # order.
-            for i in (4*y + x for x in range(4) for y in range(4)):
+            for i in (4*x + y for y in range(4) for x in range(4)):
                 subregion = self.subregions[i]
                 if subregion is not None:
                     for (region, coremask) in \
--- a/rig/machine_control/regions.py
+++ b/rig/machine_control/regions.py
@@ -12,6 +12,7 @@ selected.
 A complete introduction and specification of the region system is given in
 "Managing Big SpiNNaker Machines" By Steve Temple.
 """
+import array
 import collections
 from six import iteritems
 
@@ -51,26 +52,6 @@ def get_region_for_chip(x, y, level=3):
     return region
 
 
-def minimise_regions(chips):
-    """Create a reduced set of regions by minimising a hierarchy tree.
-
-    Parameters
-    ----------
-    chips : iterable
-        An iterable returning x and y co-ordinate pairs.
-
-    Returns
-    -------
-    generator
-        A generator which yields 32-bit region codes which minimally cover the
-        set of given chips.
-    """
-    t = RegionTree()
-    for (x, y) in chips:
-        t.add_coordinate(x, y)
-    return t.get_regions()
-
-
 def compress_flood_fill_regions(targets):
     """Generate a reduced set of flood fill parameters.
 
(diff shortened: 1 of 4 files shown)


rig/machine_control/regions.py
```diff
@@ -144,63 +117,82 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
+            raise ValueError((x, y, p))
 
         # Determine which subregion this refers to
         subregion = ((x >> self.shift) & 0x3) + 4*((y >> self.shift) & 0x3)
 
         if self.level == 3:
             # If level-3 then we just add to the locally selected regions
-            self.locally_selected.add(subregion)
+            self.locally_selected[p] |= 1 << subregion
         else:
```
mundya: Should also check if the subregion has already been selected, otherwise the same core could be specified twice when calling getting regions and core masks out.

mossblaser:
This looks reasonable to me though I think I'd better check through it again on monday if that is OK?

mundya:
> This looks reasonable to me though I think I'd better check through it again on monday if that is OK?

No worries, this is hardly an urgent change!

rig/machine_control/regions.py
```diff
@@ -144,75 +117,102 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
 
-        # Include subregions if they exist
-        if self.level < 3:
-            for i, sr in enumerate(self.subregions):
-                if i not in self.locally_selected and sr is not None:
-                    for r in sr.get_regions():
-                        yield r
+        # Order the locally selected items and then yield them
+        sorted_subregion_coremasks = sorted(list(subregions_cores.items()),
+                                            key=lambda x: (x[0] << 32) | x[1])
```
mossblaser: Why 32?

rig/machine_control/regions.py
```diff
@@ -144,75 +117,102 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
+        for (subregions, coremask) in sorted_subregion_coremasks:
+            yield (region_code | subregions), coremask
 
-    def add_coordinate(self, x, y):
-        """Add a new coordinate to the region tree.
+        if self.level < 3:
+            # Iterate through the subregions and recurse, we iterate through in
+            # the order which ensures that anything we yield is in increasing
+            # order.
```
mossblaser: I trust that SC&MP expects low-numbered levels first?

rig/machine_control/regions.py
```diff
@@ -144,75 +117,102 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
+            for i in (4*x + y for y in range(4) for x in range(4)):
+                subregion = self.subregions[i]
+                if subregion is not None:
+                    for x in subregion.get_regions_and_coremasks():
+                        yield x
```
mossblaser: Not keen on "x" here; would prefer (region, coremask) or something.

rig/machine_control/regions.py
mundya: It expects the combination `(region << 18) | coremask` in increasing orders... so yeah, low-numbered levels go first because they are the next significant field after y and x.

rig/machine_control/regions.py
```diff
@@ -144,75 +117,102 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
-
-        # If "full" then return True (i.e., would be 0x____ffff if converted)
-        return self.locally_selected == {i for i in range(16)}
+                self.subregions[subregion] = RegionCoreTree(base_x, base_y,
+                                                            self.level + 1)
+
+            # If the subregion reports that all of its subregions for this core
+            # are selected then we need to add it to `locally_selected`.
+            if self.subregions[subregion].add_core(x, y, p):
+                self.locally_selected[p] |= 1 << subregion
```
mossblaser: Should you not also remove the subregion from `self.subregions` in this case?

rig/machine_control/regions.py
mundya: This could be:

```python
... = sorted(subregions_cores.items())
```

With no loss of correctness, I'll change that.

rig/machine_control/regions.py
```diff
@@ -144,75 +117,102 @@ def __init__(self, base_x=0, base_y=0, level=0):
...
+                subregions_cores[subregions] |= 1 << core
 
-        # Include subregions if they exist
-        if self.level < 3:
-            for i, sr in enumerate(self.subregions):
-                if i not in self.locally_selected and sr is not None:
-                    for r in sr.get_regions():
-                        yield r
+        # Order the locally selected items and then yield them
+        sorted_subregion_coremasks = sorted(list(subregions_cores.items()),
```
mossblaser: six.iteritems

mundya: In this case, six and two threes.

mossblaser:
Right; aside from the "x" comment; this looks correct to me, even if I can't quite work out why it should be better than the previous technique per-se. Feel free to merge at your discretion.

mundya:
This should be ready for re-review or merge, up to you @mossblaser

mossblaser:
Aside from the above question, this LGTM!

PR #211, opened 2015-12-16
Routing table tools

mundya:
Adds some tools for working with routing tables.

 - [x] Method for finding common `X`s in routing table entries
 - [x] Methods for expanding routing table entries to remove `X`s
 - [x] Tool for checking whether one routing table is a functional subset of another
 - [x] Routing table minimisation
 - [x] Add to documentation
 - [x] Update SC&MP to return routing-table availability in system info
 - [x] Update ChipInfo object to include routing table usage
 - [x] Add utility function in ~~routing_table.utils~~ place_and_route.utils which converts from SystemInfo to table-availability lookup.
 - [x] Integrate into P&R wrapper(?)
 - [x] Merge in #213 with improvements to this API


rig/routing_table/minimise.py
```diff
@@ -0,0 +1,454 @@
...
+        routing_table,
+        key=lambda entry: get_generality(entry.key, entry.mask)
+    )
+
+    while len(routing_table) > target_length:
+        # Get the best merge
+        merge = get_best_merge(routing_table, aliases)
+
+        # If there is no merge then stop
+        if merge.goodness < 0:
```
mundya: This should be `if merge.goodness <= 0`

mundya:
@mossblaser I think this is ready for review/would benefit from your thoughts. Phase 2 of routing table minimisation may come along in a later PR.

mundya:
@mossblaser - I'm tempted to rewrite the minimiser to allow it to work with routing trees as well so that I can make use of default entries.

mundya:
Some stats for anyone following this:

Table name | # entries | # unique routes | Time to minimise (target length=0) / s | # minimised entries
---------- | --------- | --------------- | -------------------------------------- | -------------------
pynn1 | 2005 | 220 | 14.74 | 566
pynn2 | 2007 | 210 | 15.98 | 527
nengo_1739 | 1738 | 28 | 0.54 | 38

These are just some routing tables that I had lying around - I'll go hunting for a larger dataset when this floats back to the top of my priority queue.

That said, this appears to minimise routing tables somewhat effectively.

mundya:
With `target_length=1023` rather than `0`.

Table name | # entries | # unique routes | Time to minimise (target length=1023) / s | # minimised entries
---------- | --------- | --------------- | -------------------------------------- | -------------------
pynn1 | 2005 | 220 | 2.71 | 1017
pynn2 | 2007 | 210 | 3.40 | 1013
nengo_1739 | 1738 | 28 | 0.29 | 955

mossblaser:
> Scrap that, for the moment I'm going to argue that the likelihood of using default entries having any serious benefit when also using minimisation is slim.

I'd certainly be interested in seeing some "ball-park" numbers as to what you lose from not using default routes. I vaguely remember the southampton guys mentioning this sort of thing too...

> These are just some routing tables that I had lying around - I'll go hunting for a larger dataset when this floats back to the top of my priority queue.

I think this would be worth a look for more up-to-date nengo things since you've changed a lot of late...

> That said, this appears to minimise routing tables somewhat effectively.

Yes indeed! Particularly like the performance numbers when you just say "stop at 1024". On this note we *should* really work out what this number is by asking the machine... It does seem to me, though, that there is little incentive to do so since "sharing" a routing table is such a liability...

On this note, would it be worth including this tool as part of the P&R wrapper?

rig/place_and_route/utils.py
```diff
@@ -277,6 +277,10 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
         If a routing tree has a terminating vertex whose route is set to None,
         that vertex is ignored.
 
+    .. note::
+        :py:func:`~.build_and_minimise_routing_tables` can be used to get
```
mossblaser: "can be used to minimise routing tables produced by this function."

I'd include an example too (especially if you need to turn off default-route omission!)

rig/place_and_route/utils.py
```diff
@@ -360,6 +364,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+    attempt to minimise those tables to meet a target length.
+
+    This is a wrapper around :py:func:`~.build_routing_tables` and
+    :py:func:`rig.routing_table.minimise`.
+
+    .. warning::
+        It may not be possible to minimise all tables such that they are
+        shorter than `target_length`. The chip co-ordinates of any routing
+        tables that could not be sufficiently minimised are reported as the
+        second returned value.
```
mossblaser: I'd imagine that any application which doesn't meet the target given would have no option at this point but to bail out. As such I'd suggest just making this throw an exception instead.

rig/place_and_route/utils.py
```diff
@@ -360,6 +364,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+
+    Parameters
+    ----------
+    routes : {net: :py:class:`~rig.place_and_route.routing_tree.RoutingTree`, \
+              ...}
+        The complete set of RoutingTrees representing all routes in the system.
+        (Note: this is the same datastructure produced by routers in the `par`
```
mossblaser: Actually the reference to a `par` module is certainly a stale comment from a *long* time ago. This should refer to the "rig.place_and_route" module. Would you mind fixing the same error in the other table building utility?

mundya:
> I'd certainly be interested in seeing some "ball-park" numbers as to what you lose from not using default routes. I vaguely remember the southampton guys mentioning this sort of thing too...

I have no numbers, but as this algorithm generates entries with large numbers of Xs my gut feeling is that even *if* a routing entry which could be default routed wasn't merged with anything else you probably wouldn't be able to remove it from the table anyway.

> I think this would be worth a look for more up-to-date nengo things since you've changed a lot of late...

Again, no numbers, but this is sufficiently effective that the larger circular convolution models actually fit on the machine.

> On this note we should really work out what this number is by asking the machine... It does seem to me, though, that there is little incentive to do so since "sharing" a routing table is such a liability...

I'd agree that sharing routing tables is a Bad Idea (:tm:) unless you can guarantee that the entries which you have to share with are higher priority than the entries you want to minimise *and* that the unminimised table (including the other entries) is orthogonal.

+1 for asking the machine how any entries are available! Hard-coding 1023 makes me feel very unhappy.

> On this note, would it be worth including this tool as part of the P&R wrapper?

I'm happy to do that provided we have a flag to turn it off and document thoroughly... but of course these are things you'd do anyway ;)

rig/place_and_route/utils.py
mundya: I swithered on this, but yeah, I'm happy to raise an exception. Shall change.

rig/place_and_route/utils.py
mundya: Shall do :) Copy-pasta is fun!

mossblaser:
> Again, no numbers, but this is sufficiently effective that the larger circular convolution models actually fit on the machine.

Already got a nengo prototype running? Super; I'd call that better than numbers ;).

> +1 for asking the machine how any entries are available! Hard-coding 1023 makes me feel very unhappy.

How should this information propagate? Via the P&R infrastructure (e.g. constraints) or via a side channel (e.g. directly passing in a system-info or a raw dictionary?)

rig/place_and_route/utils.py
```diff
@@ -360,6 +364,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+    -------
+    {(x, y): [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+        Minimised routing tables. The correct operation of these tables relies
+        on their entries not being reordered.
+    {(x, y), ...}
+        Co-ordinates of chips for which tables shorter than `target_length`
+        could not be produced.
+    """
+    # Build the tables and then minimise them
+    tables = {
```
mossblaser: Due to the CPU-bound nature of the problem and trivial parallelism available here it may be worth just dropping in: https://docs.python.org/2/library/multiprocessing.html#using-a-pool-of-workers as a quick and easy speedup.

rig/place_and_route/utils.py
```diff
@@ -360,6 +364,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+    Parameters
+    ----------
+    routes : {net: :py:class:`~rig.place_and_route.routing_tree.RoutingTree`, \
+              ...}
+        The complete set of RoutingTrees representing all routes in the system.
+        (Note: this is the same datastructure produced by routers in the `par`
+        module.)
+    net_keys : {net: (key, mask), ...}
+        The key and mask associated with each net.
+    target_length : int
```
mossblaser: This probably needs to be (at least optionally) some sort of lookup from (x, y) to number of entries really...

mundya:
> How should this information propagate? Via the P&R infrastructure (e.g. constraints) or via a side channel (e.g. directly passing in a system-info or a raw dictionary?)

I think it feels somewhat similar to the SDRAM constraint, so I guess via the P&R infrastructure... But I have no strong preference. I was secretly hoping you'd push the API for this into shape ;)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+
+We need a set of rules for:
+
+ 1. Where merged entries are to be inserted into the table
+ 2. Which merges are allowed
+
+"Ordered Covering"
+==================
+
+The algorithm implemented here, "Ordered Covering", provides the following
```
mossblaser: If we're likely to have multiple algorithms available for minimisation (just as we have them available for P&R) it may be a good idea to rename this file to reflect the name of the algorithm and then provide an alias (as you have done) in the top level. Will the different algorithms be sufficiently overlapping in compatibility that this would be a reasonable idea?

mundya: Yes, this is probably a reasonable idea.

mossblaser:
>> How should this information propagate? Via the P&R infrastructure (e.g. constraints) or via a side channel (e.g. directly passing in a system-info or a raw dictionary?)

>I think it feels somewhat similar to the SDRAM constraint, so I guess via the P&R infrastructure... But I have no strong preference. I was secretly hoping you'd push the API for this into shape ;)

Yeah, the other difficulty is that no vertex is realistically going to consume routing table entries so putting this as a resource seems slightly awkward. Doing it via constraints also seems a good idea but only if this module were a part of the P&R module. Since I'm still unsure whether this belongs in P&R or separately... Doing it via a dictionary seems like by-far the cleanest solution when you look at the function in isolation though...

I'll continue to think about this...

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+rules:
```
mossblaser: If I understand correctly, these rules are written from the perspective of something *generating* routing table entries sequentially(?) rather than as a property of a complete routing table. This may be worth stating explicitly.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+
+ 1. Routing tables are to be kept sorted in increasing order of "generality",
+    that is the number of ``X``s in the entry. An entry with the key-mask pair
+    ``00XX`` must be placed below any entries with fewer ``X``s in their
+    key-mask pairs (e.g., below ``0000`` and ``000X``).
```
mossblaser: This is a partial ordering, right? (e.g. 0X00 and 00X0 have no ordering relations).

mundya: Yes.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+Following these rules ensures that the minimised table will be functionally
+equivalent to the original table provided that the original table was invariant
+under reordering OR was provided in increasing order of generality.
+"""
+from collections import defaultdict, namedtuple
+from rig.routing_table import RoutingTableEntry
+from six import itervalues
+
+
+def minimise(routing_table, target_length=1023):
```
mossblaser: I'd say that target_length should default to 1024: If an assumption about table availability is to be made it should be made by end users, not in Rig. Hopefully we'll come up with something to solve this problem...

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+    ----------------
+    target_length : int
+        Target length of the routing table; the minimisation procedure will
+        halt once either this target is reached or no further minimisation is
+        possible.
+
+    Returns
+    -------
+    [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+        Reduced routing table entries.
```
mossblaser: As with the wrapper, *maybe* this should raise an exception if the target isn't met? Not many applications will have any sensible course of action to take when minimisation fails and forcing everyone to manually check the output length seems like unnecessary boilerplate.

rig/routing_table/minimise.py
mundya: I'm not sure I'd use the term "*generating*", but if the table you receive is in order of increasing generality and you apply these rules when you remove and add entries to the table then the final table, and every intermediate table, is guaranteed to be functionally equivalent to the starting table.

(*Note* if the starting table is ordered but is non-orthogonal then following these rules will still guarantee a functionally equivalent table but it may not be a small as it could be if you removed the non-orthogonal entries.)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+    """
+    def get_best_merge(routing_table, aliases):
```
mossblaser: Does this really *need* to be a closure?

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+ Key       Mask      Route
+ =======   ========  =================
+ ``0000``  ``1111``  North, North East
+ ``0111``  ``0111``  South
+ =======   ========  =================
+
+Which, from now on, will be written as::
+
+    0000 -> N NE
+    X111 -> S
```
mossblaser: Might be nice to make the RoutingTableEntry `__str__` produce this format(?)

mundya: Totally up for that!

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+    merge : :py:class:`~.Merge`
+        Initial merge to refine.
+    aliases : {(key, mask): {(key, mask), ...}, ...}
+        Map of key-mask pairs to the sets of key-mask pairs that they actually
+        represent.
+
+    Yield
+    -----
+    :py:class:`~.Merge`
+        New set of entries which may be merged.
```
mossblaser: It appears you rely on the "goodness" of subsequently produced merges being monotonically decreasing. If so, this should be documented.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+
+class Merge(namedtuple("Merge", ["routing_table", "entries", "key", "mask",
+                                 "generality", "goodness",
+                                 "insertion_index"])):
+    """Represents a potential merge of routing table entries.
+    """
+    def __new__(cls, routing_table, entries=set()):
+        # Generate the new key and mask
+        ones = 0x00000000  # Wherever there is a 1 in *any* of the keys
+        key = 0xffffffff  # ... 1 in *all* of the keys
```
mossblaser: The above names don't seem that helpful(?)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+           gg(routing_table[pos]) <= generality):
+        pos += 1
+
+    return pos
+
+
+class Merge(namedtuple("Merge", ["routing_table", "entries", "key", "mask",
+                                 "generality", "goodness",
+                                 "insertion_index"])):
+    """Represents a potential merge of routing table entries.
```
mossblaser: Could do with attribute documentation!

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+            # Get the entry
+            entry = routing_table[i]
+
+            # Update the values
+            ones |= entry.key
+            key &= entry.key
+            mask &= entry.mask
+
+        # Compute the new mask, key and generality
+        # NOTE: `~key` indicates wherever there is a `0` in any key.
```
mossblaser: Just create `zeros = ~key` then ;).

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+        # Consequently `ones ^ ~key` indicates any bits in which there are
+        # either only zeros or ones (or, the bits where there doesn't need to
+        # be an X in the new entry).
```
mossblaser: Create `fixed = ones ^ zeros`?

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+    def apply(self, aliases):
+        """Apply the merge to the routing table it is defined against and get a
+        new routing table and alias dictionary.
+
+        Returns
+        -------
+        [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+            A new routing table which may be minimised further.
+        {(key, mask): {(key, mask), ...}}
+            A new dictionary mapping keys and masks from the new routing table
```
mossblaser: "A new aliases dictionary"(?)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+            route=routing_entries[next(iter(self.entries))].route,
+            key=self.key, mask=self.mask
+        )
+
+        routing_entries.insert(self.insertion_index, new_entry)
+        aliases[(self.key, self.mask)] = our_aliases = set([])
+
+        # Remove all the old entries
+        offset = 0
+        for i in sorted(self.entries):
```
mossblaser: As I understand this relies on the fact that all merged entries must have lower generalities and thus be higher-up the table than the entry we just inserted meaning the insertion hasn't broken the indexes of the entries we care about? This should probably be explicitly pointed out.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+
+        Returns
+        -------
+        [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+            A new routing table which may be minimised further.
+        {(key, mask): {(key, mask), ...}}
+            A new dictionary mapping keys and masks from the new routing table
+            to sets of keys and masks that they represent.
+        """
+        # Generate a new empty table of the correct size
```
mossblaser: empty?

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,557 @@
...
+        routing_entries = list(self.routing_table)
+        aliases = dict(aliases)
+
+        # Add the new entry
+        new_entry = RoutingTableEntry(
+            route=routing_entries[next(iter(self.entries))].route,
+            key=self.key, mask=self.mask
+        )
+
```
mossblaser: Given the above comment, I'd remove this empty line.

rig/routing_table/minimise.py
mundya: Comment out of sync error...

rig/routing_table/minimise.py
mundya: I've switched these to `any_ones` and `all_ones`, and have changed `mask` to `all_set`.

rig/routing_table/minimise.py
mundya: Note: added as `any_zeros_or_xs`

rig/routing_table/minimise.py
mundya: Exchanging lower and higher, but yes, I'll add a comment to that effect.

rig/routing_table/minimise.py
mundya: And is done.

rig/routing_table/minimise.py
mundya: Fixed.

rig/routing_table/minimise.py
mundya: Done.

rig/routing_table/minimise.py
mundya: Done.

rig/routing_table/minimise.py
mundya: Some similar done.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+class Merge(namedtuple("Merge", ["routing_table", "entries", "key", "mask",
+                                 "generality", "goodness",
+                                 "insertion_index"])):
+    """Represents a potential merge of routing table entries.
+    """
+    def __new__(cls, routing_table, entries=set()):
+        # Generate the new key and mask
+        any_ones = 0x00000000  # Wherever there is a 1 in *any* of the keys
+        all_ones = 0xffffffff  # ... 1 in *all* of the keys
+        all_set = 0xffffffff  # ... 1 in *all* of the masks
```
mossblaser: `all_selected`(?) "set" doesn't imply something "masky" to me?

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+            route=routing_entries[next(iter(self.entries))].route,
+            key=self.key, mask=self.mask
+        )
+
+        routing_entries.insert(self.insertion_index, new_entry)
+        aliases[(self.key, self.mask)] = our_aliases = set([])
+
+        # Remove all the old entries
+        offset = 0
+        for i in sorted(self.entries):
```
mossblaser: As noted in old version, an explicit note that the new entry insertion is guaranteed to be after all removed entries (and thus can't make their indexes wrong) would be a welcome reassurance.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+        # Generate a new empty table of the correct size
+        routing_entries = list(self.routing_table)
+        aliases = dict(aliases)
+
+        # Add the new entry
+        new_entry = RoutingTableEntry(
+            route=routing_entries[next(iter(self.entries))].route,
+            key=self.key, mask=self.mask
+        )
+
```
mossblaser: I'd remove the blank line to make the comment above obviously associated with the following two lines.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+    .. warning::
+
+        Intermediate stages are yielded explicitly so that the refinement
+        process can be stopped early. Only the last yielded merge will not
+        alter the behaviour of the routing table when applied.
+    """
+    # We yield the merge as it is so that an initial test for goodness can
+    # be made.
+    yield merge
```
mossblaser: What guarantees that this is a valid merge?

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+
+    # Perform the down-check
+    for merge in refine_downcheck(merge, aliases):
```
mossblaser: That this modifies the "merge" variable is a little scary: needs a comment stating that this is intentional.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+        yield merge
+
+    # Perform the up-check
+    changed = False
+    for merge in refine_upcheck(merge):
```
mossblaser: Again, a comment explaining that "merge" is changed is needed.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+
+            # Add to the alias list
+            our_aliases.update(aliases.pop(km, set([km])))
+
+        return routing_entries, aliases
+
+
+def refine_merge(merge, aliases):
+    """Remove entries from a merge to generate new merges which meet the rules
+    required to allow them to be inserted into the routing table.
```
mossblaser: Right, at long last I've finally understood what's going on with this function(!)

This function iteratively removes key/mask pairs from a candidate merge until the merge does not violate the ordered covering rules. In order to allow early termination of this process when the merge candidate becomes inferior to some other candidate merge, it is implemented as a generator which produces "half-finished" merges of monotonically decreasing goodness. The user of the generator is then able to terminate the generator as soon as the refined merges drop below a particular quality level.

Now that I understand this, I would argue that this is not the clearest way of implementing this. I think this function should take a minimum goodness as an argument and terminate itself (returning None or something) hopefully making the implementation much simpler.

Since (skipping ahead) I see that you're testing intermediate refinements, the best way to do this may be to rename this `_refine_merge` and make a wrapper `refine_merge` which does what I described above.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+            entry = routing_table[i]
+
+            # Update the values
+            any_ones |= entry.key
+            all_ones &= entry.key
+            all_set &= entry.mask
+
+        # Compute the new mask, key and generality
+        any_zeros_or_xs = ~all_ones
+        mask = all_set & (any_ones ^ any_zeros_or_xs)
```
mossblaser: Could you add (yet another) intermediate step:

`changes_or_x = any_ones ^ any_zeros_or_xs`

As I understand this gives the set of bits which are either already X or which we have examples of both 1 and 0. This just makes it clearer that we're adding Xs to the mask for which we know we need to match both 0 and 1.

As an aside: the fact that there are 1s in any_zeros_or_xs and changes_or_x when we have an X really doesn't matter since all this only ends up getting ANDed with the mask in the end anyway. While I appreciate that not including the "or_x" bits of these names is incorrect (and possibly going to worry a careful reader), including them will also worry a potential reader looking for why it matters (see: me!). I'm not sure whether this justifies a short comment to reassure or not...

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+        if any(overlap(key, mask, other.key, other.mask) for other in
+               merge.routing_table[i+1:merge.insertion_index]):
+            # The entry would be partially or wholly covered by another entry,
+            # remove it from the merge and yield a new merge.
+            merge = Merge(merge.routing_table, merge.entries - {i})
+            yield merge
+
+
+def refine_downcheck(merge, aliases):
+    """Shrink the merge to avoid it covering up any entries which are below the
```
mossblaser: s/Shrink/prune/

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,608 @@
...
+
+        .. note::
+
+            Use :py:meth:`~rig.routing_table.expand_entries` to generate an
+            orthogonal table and receive warnings if the input table were not
+            orthogonal.
+
+    .. warning::
+
+        The input routing table *must* also include entries which could be
```
mossblaser: This should really be the first warning. I suspect most people will be producing orthogonal tables without thinking (if I understand rightly, the routing table generation function already does this, right?) but omitting default routes is a really easy way to very confusingly shoot yourself in the foot since the results can clearly work *some* of the time...

mundya: > (if I understand rightly, the routing table generation function already does this, right?)

It's a function of the keys and masks you use; I don't think `build_routing_tables` does anything at all that guarantees anything about the orthogonality or not of the resulting tables. In fact, if you did assign your keys such that it produced a non-orthogonal table you could really shoot yourself in the foot as it does nothing to order to tables particularly.

rig/routing_table/ordered_covering.py
mundya: `all_not_xs`?

mundya: `none_xs`?

rig/routing_table/ordered_covering.py
mundya: If none of the checks yield anything at all then this must be valid, otherwise it won't be the last merge that's yielded and should be disregarded anyway.

rig/routing_table/ordered_covering.py
mundya: Note: if you're using one `BitField` to assign *all* your keys then you are guaranteed orthogonal routing tables.

mundya: ...and this should probably be mentioned somewhere...

mundya:
Just to confirm where we are:

 - `build_and_minimise_routing_tables` needs to:
   - [ ] also accept a dictionary mapping co-ordinate to maximum table length
   - [ ] raise an exception if any table is too long
 - [ ] correct a documentation bug in both `build_routing_tables` and `build_and_minimise...`

Is that all?

mossblaser:
Sorry, I was interrupted with my review yesterday so I'll resume then;
still not reviewed the up/down functions... Sorry!

Jonathan

mundya:
Could we add number of available routing table entries to the `SystemInfo` object?

rig/place_and_route/utils.py
mundya: Done.

mossblaser:
> Could we add number of available routing table entries to the `SystemInfo` object? I guess actually the `ChipInfo`s.

That is my plan. I'd suggest we get that into this PR so we don't have to change the API later... I'm afraid time is against me getting that done at short notice though, sorry!

mundya:
> That is my plan. I'd suggest we get that into this PR so we don't have to change the API later...

That's good by me. If I have spare time and you don't beat me to it I'll look at what needs doing in SCAMP for this.

mossblaser:
If you want to do this I'd suggest branching your patch off 1392251
<https://bitbucket.org/spinnaker-low-level-software/spinnaker_tools/commits/1392251effded9541978aa8f75c1a5475b06af88>
since
later revisions add features which change some entries in the sark struct
which we should postpone for the formal 2.0.0 release of the tools.
Hopefully won't be huge amounts of work!

rig/routing_table/errors.py
```diff
@@ -0,0 +1,16 @@
+class MinimisationFailedError(Exception):
```
mossblaser: I realise it may be the wrong convention but elsewhere in rig files containing just exceptions have been called "exceptions.py". Might be worth keeping things consistent here if you don't mind?

mundya: Yeah, that's fine, I should've checked, sorry.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+        1000 -> N
+        X000 -> NE
+
+    Cannot be merged with the first two entries because that would move it
+    below the entry with the key-mask pair of ``X000``, which would cover it.
+
+    Returns
+    -------
+    :py:class:`~.Merge`
+        New merge with entries possibly removed.
```
mossblaser: "If the goodness of the solution is discovered to be less or equal to ``min_goodness``, the function will terminate early and an empty merge will be returned."

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+            # If the up-check removed any entries we need to re-perform the
+            # down-check; but we do not need to re-perform the up-check as the
+            # down check can only move the resultant merge nearer the top of
+            # the routing table.
+            merge = refine_downcheck(merge, aliases, min_goodness)
+
+    return merge
+
+
+def overlap(key_a, mask_a, key_b, mask_b):
```
mossblaser: Docstring with some statement as to what overlapping means, please :)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+
+    For example, the third entry of::
+
+        0011 -> N
+        0100 -> N
+        1000 -> N
+        X000 -> NE
+
+    Cannot be merged with the first two entries because that would move it
+    below the entry with the key-mask pair of ``X000``, which would cover it.
```
mossblaser: Being hyper-picky it would be worth explicitly stating that the merge would give you a key `XXXX` or something.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+    also remove it.  For this example we would then consider merging only the
+    first two entries, leading to a new key-mask pair of ``000X1`` which can be
+    safely inserted between ``00X00`` and ``XX1XX``::
+
+        00100 -> N S
+        00X00 -> N S
+        000X1 -> N S
+        XX1XX -> 3 5
+
+    *Note*: If the original table were orthogonalised to::
```
mossblaser: I presume the reader should interpret this note as being a hint that you should definitely orthogonalise your routing tables before beginning? May be worth stating explicitly.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+    The table could be reduced to::
+
+        00100 -> N S
+        000XX -> N S
+        XX1XX -> 3 5
+
+    Returns
+    -------
+    :py:class:`~.Merge`
+        New merge with entries possibly removed.
```
mossblaser: "If the goodness of the solution is discovered to be less or equal to ``min_goodness``, the function will terminate early and an empty merge will be returned."

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+    """
+    def get_aliased_keys_and_masks(merge):
```
mossblaser: I always end up slightly nervous when I encounter a non-trivial nested function like this expecting it to be a closure and have some side-effects. It would set my mind at ease if it was broken out and _prefixed or something but that may just be me being too picky.

mundya: Yeah, that's fine.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+
+    while merge.entries and merge.goodness > min_goodness:
+        aliased = list(get_aliased_keys_and_masks(merge))
+
+        # If there are no aliased entries then we're done
+        if not aliased:
+            break
+
+        # Set of bit positions
+        all_bits = {1 << i for i in range(32)}
```
mossblaser: This should probably be a list; I don't see any reason to use a set(?)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+        `target_length`.
+
+    Returns
+    -------
+    [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+        Reduced routing table entries.
+    """
+    # We maintain a dictionary of which keys and masks in the routing table are
+    # combinations of other keys and masks; this allows us to be less
+    # aggressive in determining whether inserting a new entry into the table
```
mossblaser: "We maintain a dictionary of which keys and masks in the routing table are combinations of other (now removed) keys and masks; this allows us to consider only the keys and masks the user actually cares about when determining if inserting a new entry will break the correctness of the table."

rig/routing_table/ordered_covering.py
mundya: I guess a tuple'd be fine actually.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+        00001 -> N S
+        00011 -> N S
+        00100 -> N S
+        00X00 -> N S
+        XX1XX -> 3 5
+
+    Merging the first four entries would generate the new key-mask ``00XXX``
+    which would be inserted above the entry with the key-mask ``XX1XX``.
+    However ``00XXX`` would stop the key ``00110`` from reaching its correct
+    route. To avoid this one could just abandon the merge entirely, but a
```
mossblaser: "However ``00XXX`` would stop the key ``00110`` from reaching its correct route, that is, ``00110`` would be aliased by ``00XXX``."  -- Just to introduce the term used later in the code. (I hope I got that the right way around...)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+    better solution is to attempt to reduce the merge such that it no longer
+    covers any entries below it.
```
mossblaser: s/covers/aliases/ ?

mundya: I'm very bad with terminology. Combined with the aliases being a verb and a noun (pl)...

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+
+    To do this we first identify the bits that ARE ``X`` s in the generated
```
mossblaser: s/generated/merged/

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+    key-mask but which are NOT ``X`` s in the entry that we're covering. For
```
mossblaser: s/covering/aliasing/

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,635 @@
...
+                        options[(bit, False)].add(i)
+                        options[(bit, True)].add(i)
+                    elif bool(bit & entry.key) is val:
+                        # If there is a 0 in this position and we want a 0 (val
+                        # is True) then mark this entry for removal. Likewise
+                        # if there is a 1 in this position and we want a 1.
+                        options[(bit, val)].add(i)
+
+            # Get the smallest set of entries to remove and remove them
+            choice = sorted(itervalues(options), key=len)[0]
```
mossblaser: sorted(...)[0]  --> min(...)

mundya: Uh... that's embarrassing :(

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    # ---------
+    # While the merge is still better than `min_goodness` we determine which
+    # entries below it in the table it covers. For each of these covered
+    # entries we find which bits are Xs in the merged entry and are NOT Xs in
+    # the covered entry.
+    #
+    # For example:
+    #
+    #     Merged entry:      ...0XXX1...
+    #     Covered entry:     ...010XX...
```
mossblaser: Surely 010XX cannot be covered by 0XXX1 since it has fewer Xs?

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    # In this case there are 2 ``settable`` bits of interest highlighted.
+    # Whenever we have multiple covered entries we care only about the entries
+    # with the fewest number of ``settable`` bits because these most constrain
+    # which entries we may remove from the merge to avoid covering up the lower
+    # entry.
+    #
+    # NOTE:
+    #   * If there is only 1 ``settable`` bit then we are very constrained in
+    #     terms of which entries must be removed from the merge to avoid
+    #     covering a lower entry.
```
mossblaser: "This is ideal since fewer entries are likely to be removed to meet such a constraint." (?)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    # the covered entry.
+    #
+    # For example:
+    #
+    #     Merged entry:      ...0XXX1...
+    #     Covered entry:     ...010XX...
+    #     Bits of interest:      ^^
+    #     Label used below:      mn
+    #
+    # In this case there are 2 ``settable`` bits of interest highlighted.
```
mossblaser: Could you explicitly define ``settable`` here?

rig/routing_table/ordered_covering.py
mundya: 010XX could be part of some other entry (as in, XXXXX could contain it, so the aliases dict would have `{XXXXX: {010XXX}}`. I'll point that out in the comment.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    #
+    # NOTE:
+    #   * If there is only 1 ``settable`` bit then we are very constrained in
+    #     terms of which entries must be removed from the merge to avoid
+    #     covering a lower entry.
+    #   * If there are no ``settable`` bits then we cannot possibly avoid
+    #     covering the lower entry - the only correct action is to return an
+    #     empty merge.
+    #
+    # Assuming that there were no covered entries with no ``settable`` bits
```
mossblaser: "Assuming that there were no covered entries **without any** ``settable`` bits" ??

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    #     options = {(m, True): {1, 4, 5},
+    #                (n, False): {3, 7}}
+    #
+    # Indicating that we'd need to remove entries 1, 4 and 5 from the merge to
+    # "set" the mth bit of the merged to 0 or that we'd need to remove entries
+    # 3 and 7 to set the nth bit of the merged entry to set the nth bit to 1.
+    #
+    # NOTE:
+    #   The boolean part of the pair indicates which value needs to be removed
+    #   (True -> remove all 1s; False -> remove all 0s). If all Xs and 1s in a
```
mossblaser: "remove all 1s **and Xs**" and "remove all 0s **and Xs**"? Might be clearer than stating it separately later.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    #   given bit position are removed from a merge then the merged entry is
+    #   guaranteed to have a 0 in the bit position. Vice-versa removing all Xs
+    #   and 0s in a given bit position from a merge will result in a merged
+    #   entry with a 1 in that position.
+    #
+    # As we want to make our merges as large as possible we select the smallest
+    # set of entries to remove from the merge from ``options``. After removing
+    # these entries we have a new merge with lower generality.
```
mossblaser: You can remove this last sentence, you re-state this fact at a more pertinent time in the next paragraph.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    #
+    # For two reasons:
```
mossblaser: "The whole process is then repeated since:" (also replaces the tail-end of the sentence after the bullets).

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+    # there are then we resolve the covering as above, if there aren't then (as
+    # far as the down-check is concerned) we have a valid merge so we return
+    # the merge.
+
+    # Set of bit positions
+    all_bits = tuple(1 << i for i in range(32))
+
+    # While the merge is still worth considering continue to perform the
+    # down-check.
+    while merge.goodness > min_goodness:
```
mossblaser: If the merge goodness drops below min_goodness this function should return an empty merge (as per the docstring). I can't see any code which would cause this to happen. (I *can* see code for returning empty merges when there are no settable bits, though.)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+                # Add this settable mask and the required values to the
+                # settables list.
+                bits_and_vals.update((bit, bool(key & bit)) for bit in
+                                     all_bits if bit & settable)
+
+        if most_stringent == 0:
+            # If are there any instances where we could not possibly change a
+            # bit to avoid aliasing an entry we'll return an empty merge and
+            # give up.
+            merge = Merge(merge.routing_table, set())
```
mossblaser: may as well break now?

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
+import warnings
+from rig.routing_table import RoutingTableEntry
+
+
+def table_is_subset_of(entries_a, entries_b):
+    """Check that every key matched by every entry in one table results in the
+    same route when checked against the other table.
```
mossblaser: Is it worth pointing out that this function is likely only of use for testing? (i.e. regular programs shouldn't need it?) Is it worth considering putting it in a utils module in the test suite rather than the main API?

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+        ...     RoutingTableEntry({Routes.east}, 0x1, 0xf),
+        ...     RoutingTableEntry({Routes.south_west}, 0x5, 0xf),
+        ...     RoutingTableEntry({Routes.north, Routes.north_east}, 0x8, 0xf),
+        ...     RoutingTableEntry({Routes.east}, 0x9, 0xf),
+        ...     RoutingTableEntry({Routes.south_west}, 0xe, 0xf),
+        ...     RoutingTableEntry({Routes.north, Routes.north_east}, 0xc, 0xf),
+        ...     RoutingTableEntry({Routes.south, Routes.south_west}, 0x0, 0xb),
+        ... ]
+
+    Is a functional subset of a minimised version of itself::
```
mossblaser: This sentence doesn't make sense.

rig/routing_table/__init__.py
```diff
@@ -0,0 +1,4 @@
+from rig.routing_table.entries import RoutingTableEntry, Routes
+from rig.routing_table.exceptions import MinimisationFailedError
+from rig.routing_table.ordered_covering import minimise
```
mossblaser: A comment stating that this line is responsible for choosing the "default" minimiser would be good.

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+            if (entry.mask & key) == entry.key:
+                return entry.route
+        raise NoMatchError  # No match was found at all
+
+    # Check that all the routes from the expanded first table are equivalent in
+    # the second table. NB: the first table is expanded in all positions save
+    # for where there are common Xs in the second table.
+    common_xs = get_common_xs(entries_b)
+    try:
+        return all(get_first_match(entry.key) == entry.route for entry in
```
mossblaser: As much as I like this style; this algorithm would certainly be clearer as a for-loop over the keys with the above function simply being inlined as a nested loop.

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+                   expand_entries(entries_a, ignore_xs=common_xs))
+    except NoMatchError:
+        return False
+
+
+def expand_entry(entry, ignore_xs=0x0, entry_cls=RoutingTableEntry):
```
mossblaser: Why not just always use `RoutingTableEntry`?

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+    """Turn all Xs which are not marked in `ignore_xs` into ``0``\ s and
+    ``1``\ s.
+
+    The following will expand any Xs in bits ``1..3``\ ::
+
+        >>> entry = RoutingTableEntry(None, 0b0100, 0xfffffff0 | 0b1100)
```
mossblaser: s/None/set()/

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+
+    Yield
+    -----
+    Instances of `entry_cls`
+        Instances of `entry_cls` which represent the original routing table
+        entry but with all Xs not masked off by `ignore_xs` replaced with 1s
+        and 0s.
+    """
+    # Get all the Xs in the entry that are not common
+    unique_xs = (~entry.key & ~entry.mask) & ~ignore_xs
```
mossblaser: How about `remaining_xs`?

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+        default).
+
+    Yield
+    -----
+    Instances of `entry_cls`
+        Instances of `entry_cls` which represent the original routing table
+        entry but with all Xs not masked off by `ignore_xs` replaced with 1s
+        and 0s.
+    """
+    # Get all the Xs in the entry that are not common
```
mossblaser: s/common/ignored/

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+                yield new_entry
+
+            # Stop looking for Xs
+            break
+    else:
+        # If there are no Xs then yield the entry we were given.
+        yield entry
+
+
+def expand_entries(entries, entry_cls=RoutingTableEntry, ignore_xs=None):
```
mossblaser: Again, why can't entry_cls be assumed?

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+    entries : [:py:class:`~rig.routing_table.RoutingTableEntry`...] or similar
+        The entries to expand.
+
+    Other Parameters
+    ----------------
+    entry_cls :
+        Function to use to make new routing table entries. Must have the same
+        signature as :py:class:`~rig.routing_table.RoutingTableEntry` (the
+        default).
+    ignore_xs : int
```
mossblaser: Is this ever used? If not I'd suggest removing it since if a user supplies a mask with any non-common bits ignored the orthogonality check will not be reliable.

rig/routing_table/utils.py
```diff
@@ -0,0 +1,238 @@
...
+        ... ]
+        >>> set(expand_entries(entries)) == {
+        ...     RoutingTableEntry(0, 0b0000, 0b1111),  # 0000 -> 0
+        ...     RoutingTableEntry(1, 0b0100, 0b1111),  # 0100 -> 1
+        ... }
+        True
+
+    .. warning::
+
+        The routing table that is returned is guaranteed to be orthogonal, this
```
mossblaser: "It is assumed that the input routing table is orthogonal (i.e., there are no two entries which would match the same key). If this is not the case, any entries which are covered (i.e. unreachable) in the input table will be omitted and a warning produced. As a result, all output routing tables are guaranteed to be orthogonal."

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+
+    .. warning::
+
+        The input routing table *must* also include entries which could be
+        removed and replaced by default routing.
+
+    .. warning::
+
+        It is assumed that the input routing table is not in any particular
+        order and may be reordered into ascending order of generality without
```
mossblaser: "may be reordered into ascending order of generality **(i.e. number of Xs)**"

(just to help folk who just want to use this without learning the terminology)

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+        affecting routing correctness.  It is also assumed that if this table
+        is unordered it is at least orthogonal (i.e., there are no two entries
+        which would match the same key) -- this assumption is not checked!
```
mossblaser: "this assumption is not checked" -- neither is the first one, right?

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+
+        .. note::
+
+            If *all* the keys in the table are derived from a single instance
+            of :py:class:`~rig.bitfield.BitField` then the table is guaranteed
+            to be orthogonal.
```
mossblaser: ...and reorderable.

rig/routing_table/ordered_covering.py
```diff
@@ -0,0 +1,752 @@
...
+
+        .. note::
+
+            Use :py:meth:`~rig.routing_table.expand_entries` to generate an
+            orthogonal table and receive warnings if the input table were not
```
mossblaser: s/were/is/

tests/routing_table/test_routing_table_ordered_covering.py
```diff
@@ -0,0 +1,364 @@
...
+from rig.routing_table.ordered_covering import (
+    get_generality, get_all_merges, get_insertion_index, Merge, refine_merge,
+    minimise
+)
+
+
+@pytest.mark.parametrize(
+    "key, mask, generality", (
+        (0x0, 0x0, 32),
+        (0x1, 0x1, 31),
```
mossblaser: May be worth including example of an impossible bit?

rig/routing_table/utils.py
mundya: `table_is_subset_of` uses it.

rig/routing_table/ordered_covering.py
mundya: On reflection I note that "not checked" is the reason they're assumptions :/

tests/routing_table/test_routing_table_ordered_covering.py
mundya: As `key=0x1, mask=0x0`? All of the code assumes that it won't meet any of those. Admittedly that isn't noted anywhere...

rig/routing_table/ordered_covering.py
mundya: Fixed.

rig/routing_table/ordered_covering.py
mundya: Fixed.

rig/routing_table/ordered_covering.py
mundya: Fixed.

rig/routing_table/utils.py
mundya: Possibly, but I quite like having it handy for checking how well (or otherwise) I'm driving Espresso. I appreciate that's probably a bad reason for having it hang around though.

mundya:
Not a big change in SC&MP: https://bitbucket.org/spinnaker-low-level-software/spinnaker_tools/branch/include-rtr_free-in-cmd_info

I'll compile and try to get Rig to play along shortly.

rig/routing_table/utils.py
mossblaser: Ah... Worryingly I didn't spot that... Perhaps a warning may be in order in the docstring?

tests/routing_table/test_routing_table_ordered_covering.py
mossblaser: Would be worth either testing or documenting either way. (I'd prefer the output to be valid/sensible if such an entry exists anyway...).

rig/routing_table/utils.py
mossblaser: Fair enough! It isn't exactly a controversial function! :P

mundya:
Would it just be neater to allow "minimise" to accept a SystemInfo or a dict or an int or None for `target_length`?

mossblaser:
> Would it just be neater to allow `minimise` to accept a SystemInfo or a dict or an int or None for `target_length`?

Yes and no. Certainly would be neat but the SystemInfo construct is definitely not something I want leaking everywhere... How strong are your feelings on this matter?

mundya:
> How strong are your feelings on this matter?

Not very!

mundya:
Woohoo!

PR #213, opened 2016-01-07
Improve minimisation API

mundya:


rig/place_and_route/utils.py
```diff
@@ -278,17 +277,13 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
-        minimised versions of routing tables.
-
     Parameters
     ----------
     routes : {net: :py:class:`~rig.place_and_route.routing_tree.RoutingTree`, \
               ...}
         The complete set of RoutingTrees representing all routes in the system.
-        (Note: this is the same datastructure produced by routers in the
-        `place_and_route` module.)
+        (Note: this is the same datastructure produced by routers in the `par`
```
mundya: Yay! Going backwards!

rig/place_and_route/utils.py
```diff
@@ -299,8 +294,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+    net_keys : {net: (key, mask), ...}
+        The key and mask associated with each net.
+
+    Returns
+    -------
+    {(x, y): [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
+    """
     # Pairs of inbound and outbound routes.
-    _InOutPair = namedtuple("_InOutPair", "ins, outs")
+    InOutPair = namedtuple("_InOutPair", "ins, outs")
```
mossblaser: `_InOutPair` remains in the namedtuple name field.

rig/routing_table/utils.py
```diff
@@ -1,6 +1,55 @@
 import warnings
 
-from rig.routing_table import RoutingTableEntry
+from rig.routing_table import RoutingTableEntry, MinimisationFailedError
+
+
+def remove_default_routes(table, target_length):
```
mossblaser: I would prefer to present this as a routing-table minimisation algorithm. This would mean putting it in a module named after the algorithm and calling it "minimise". Would you be OK with that? (It already meets other API requirements, I think).

rig/place_and_route/utils.py
```diff
@@ -299,8 +294,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
     -------
     {(x, y): [:py:class:`~rig.routing_table.RoutingTableEntry`, ...]
     """
+    # Build full routing tables and then remove default entries from them
```
mossblaser: A deprecation warning should be stuck here (and in the comments).

See https://github.com/project-rig/rig/blob/master/rig/machine_control/machine_controller.py#L1884-L1886
and https://github.com/project-rig/rig/blob/master/tests/machine_control/test_machine_controller.py#L2302-L2309
for how to trigger (and test for) deprecation warnings.

rig/place_and_route/utils.py
```diff
@@ -299,8 +294,54 @@ def build_routing_tables(routes, net_keys, omit_default_routes=True):
...
+            table = remove_default_routes(table, target_length=None)
+
+        # If the table is empty don't add it to the dictionary of tables.
+        if table:
+            tables[chip] = table
+
+    return tables
+
+
+def build_full_routing_tables(routes, net_keys):
```
mossblaser: I don't like this name. :( Alas, the ideal name is still probably build_routing_tables... Ugh... Any other ideas(?) I'm stuck :(.

rig/routing_table/__init__.py
```diff
@@ -1,6 +1,7 @@
 from rig.routing_table.entries import RoutingTableEntry, Routes
 from rig.routing_table.exceptions import MinimisationFailedError
-from rig.routing_table.utils import table_is_subset_of, expand_entries
+from rig.routing_table.utils import (table_is_subset_of, expand_entries,
+                                     intersect, remove_default_routes)
```
mossblaser: If `remove_default_routes` is "just another routing algorithm", there is no reason to import it into the top-level namespace.

rig/routing_table/entries.py
```diff
@@ -23,7 +24,18 @@ class RoutingTableEntry(namedtuple("RoutingTableEntry", "route key mask")):
     mask : int
         32-bit unsigned integer mask to apply to keys of packets arriving at
         the router.
+    sources : {:py:class:`~.Routes`, ...}
+        Directions in which a packet may enter the router before taking this
+        route. This is opposite to the link used (e.g., ``north`` in this set
+        indicates that a packet would enter on the ``south`` link). If the
+        `sources` and `route` are equivalent and contain only one item then it
+        *might* be possible to remove the entry and instead use default
+        routing.
```
mossblaser: "If the source routes are unknown, ``{None}`` should be used (the default)."

rig/routing_table/entries.py
```diff
@@ -23,7 +24,18 @@ class RoutingTableEntry(namedtuple("RoutingTableEntry", "route key mask")):
...
     """
+    def __new__(cls, route, key, mask, sources={None}):
+        return super(RoutingTableEntry, cls).__new__(
+                cls, route, key, mask, set(sources))
+
     def __str__(self):
         """Get an easily readable representation of the routing table entry.
```
mossblaser: Could probably do with being updated to the form "XXX...XXX -> N NE ..." when the source is {None} and "S SW -> XXX....XXX -> N NE..." when it isn't (not-showing the None in the source list, if present).

mossblaser:
Should the ordered covering implementation not update the sources set when merging? (Can't see that being done here).

mundya:
> Should the ordered covering implementation not update the sources set when merging? (Can't see that being done here).

It should, I wanted to fix the API before I played with that.

rig/routing_table/entries.py
mundya: Should the prefix indicate direction or link? (e.g., if it comes in on the south-link it's heading north, so "N -> ..." or "S -> ..."?)

mossblaser: Tricky... Gut feeling is probably this should be the name of the link the packet arrived through (not the direction of travel). After all, "Core 1" cannot be "inverted" and if something comes on on North, and West, say, it would be confusing to invert that. Thoughts?

mundya: I'm happy showing link, tbh I guess that showing direction would bleed implementation details out unnecessarily.

mossblaser:
So, aside from the name `build_full_routing_tables` which I don't like; this looks like it is about right for the API. What do you think?

We could, of course, move the routing-table generation stuff entirely into the routing_table library (i.e. including the tree->table function) which gives a little more flexibility on naming and may be the cleaner solution...

mundya:
> We could, of course, move the routing-table generation stuff entirely into the routing_table library (i.e. including the tree->table function) which gives a little more flexibility on naming and may be the cleaner solution...

I'm a fan of that idea

mundya:
- [ ] Fix some docstrings
 - [ ] Add tests for `minimise`

mundya:
**TODO**
 - [ ] Rename `ordered_covering::minimise` to `ordered_covering::ordered_covering`
  - [ ] Stop `ordered_covering::ordered_covering` from removing default entries when finished
  - [ ] Get `ordered_covering::ordered_covering` to return the aliases dict when finished

 - [ ] Add the following to `ordered_covering.py`:

```python
def minimise(routing_table, target_length):
    routing_table = ordered_covering(routing_table, target_length)
    routing_table = remove_default_routes(routing_table, target_length)
    return routing_table
```

mossblaser:
Re: 34aee59 -- superb! Absolutely the right thing to do. Also, I can't believe this wasn't the way it was done originally! (Way to write stupid code, @mossblaser...).

mundya:
@mossblaser - I think this covers everything, could I ask you to glance through and check the docs please?

docs/source/routing_table_tools_doctest.rst
```diff
@@ -120,8 +120,8 @@ minimised.
     ... ]
 
     >>> # Minimise the routing table as much as possible
-    >>> from rig.routing_table import minimise
-    >>> minimised = minimise(original, target_length=None)
+    >>> from rig.routing_table import minimise_table
```
mossblaser: The `minimise` prototype above presumably needs modifying to be called `minimise_table` to match this (sensible!) change.

mundya:
Just noticed that `table_is_subset_of` doesn't work correctly with tables that have had default routes removed.

rig/place_and_route/utils.py
```diff
@@ -257,7 +254,13 @@ def build_application_map(vertices_applications, placements, allocations,
...
 
 def build_routing_tables(routes, net_keys, omit_default_routes=True):
-    """Convert a set of RoutingTrees into a per-chip set of routing tables.
+    """**DEPRECATED** Convert a set of RoutingTrees into a per-chip set of
+    routing tables.
+
+    .. warning::
+        This method has been deprecated in favour of
+        :py:meth:`rig.routing_table.routing_tree_to_tables` and
+        :py:meth:`rig.routing_table.minimise`.
```
mossblaser: "E.g. most applications should use something like:

```py
from rig.routing_table import routing_tree_to_tables, minimise
tables = minimise(routing_tree_to_tables(routes, net_keys), target_lengths)
```

Where target_length gives the number of available routing entries on the chips in your SpiNNaker system (see :py:func:`~rig.routing_table.utils.build_routing_table_target_lengths`)"

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+from rig.routing_table import MinimisationFailedError
+from rig.routing_table.remove_default_routes import minimise as \
+    remove_default_entries
+from rig.routing_table.ordered_covering import minimise as ordered_covering
+from six import iteritems
+
+
+def minimise(routing_tables, target_lengths,
+             methods=(remove_default_entries, ordered_covering)):
+    """
```
mossblaser: Utility function which attempts to minimises routing tables for multiple chips.

For each routing table supplied, this function will attempt to use the minimisation algorithms given (or some sensible default algorithms), trying each sequentially until a target number of routing entries has been reached.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+        Dictionary mapping co-ordinates to the routing tables associated with
+        those co-ordinates. NOTE: This is the data structure as returned by
+        :py:meth:`~rig.place_and_route.build_full_routing_tables`.
+    target_lengths : int or {(x, y): int or None, ...} or None
+        Maximum length of routing tables. If an integer this is assumed to be
+        the maximum length for any table; if a dictionary then it is assumed to
+        be a mapping from co-ordinate to maximum length (or None); if None then
+        tables will be minimised as far as possible.
+    methods :
+        Iterable of minimisation methods.
```
mossblaser: Each method is tried in the order presented and the first to meet the required target length for a given chip is used. Consequently less computationally costly algorithms should be nearer the start of the list. The defaults will try to remove default routes (:py:meth:`rig.routing_table.remove_default_routes.minimise`) and then fall back on the ordered covering algorithm (:py:meth:`rig.routing_table.ordered_covering.minimise`).

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+def minimise(routing_tables, target_lengths,
+             methods=(remove_default_entries, ordered_covering)):
+    """
+
+    Parameters
+    ----------
+    routing_tables : {(x, y): [RoutingTableEntry, ...], ...}
+        Dictionary mapping co-ordinates to the routing tables associated with
+        those co-ordinates. NOTE: This is the data structure as returned by
+        :py:meth:`~rig.place_and_route.build_full_routing_tables`.
```
mossblaser: :py:meth:`~rig.routing_table.routing_tree_to_tables`. (!)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+        the maximum length for any table; if a dictionary then it is assumed to
+        be a mapping from co-ordinate to maximum length (or None); if None then
+        tables will be minimised as far as possible.
+    methods :
+        Iterable of minimisation methods.
+
+    Returns
+    -------
+    {(x, y): [RoutingTableEntry, ...], ...}
+        Minimised routing tables, guaranteed to be smaller than the table sizes
```
mossblaser: guaranteed s/to be smaller than/to be at least as small as/

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+        # Store the table if it isn't empty
+        if new_table:
+            new_tables[chip] = new_table
+
+    return new_tables
+
+
+def minimise_table(table, target_length,
+                   methods=(remove_default_entries, ordered_covering)):
+    """Apply different minimisation algorithms to minimise a table below a
```
mossblaser: I'd emphasise that this minimises a *single* routing entry just to contrast it with minimise.

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+    target length.
+
```
mossblaser: See Also
--------

minimise : Minimise routing table entries for multiple chips. (A wrapper around this function).

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+    Parameters
+    ----------
+    table : [RoutingTableEntry, ...]
+        Routing table to minimise.  NOTE: This is the data structure as
+        returned by :py:meth:`~rig.place_and_route.build_full_routing_tables`.
```
mossblaser: :py:meth:`~rig.routing_table.routing_tree_to_tables`. (!)

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+    target_length : int or None
+        Maximum length of the routing table. If None then all methods will be
+        tried and the smallest achieved table will be returned.
+    methods :
+        Iterable of minimisation methods. The defaults will try to remove
```
mossblaser: May (or may-not!) wish to use the variant of this suggested above for minimise()

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
...
+        (:py:meth:`rig.routing_table.remove_default_routes.minimise`) and then
+        apply the ordered covering algorithm
+        (:py:meth:`rig.routing_table.ordered_covering.minimise`). NOTE: methods
+        are tried in the order presented, consequently less computationally
+        costly algorithms should be nearer the start of the list.
+
+    Returns
+    -------
+    [RoutingTableEntry, ...]
+        Minimised routing table, guaranteed to be smaller than `target_length`.
```
mossblaser: guaranteed s/to be smaller than/to be at least as small as/

rig/routing_table/minimise.py
```diff
@@ -0,0 +1,112 @@
+import collections
+from rig.routing_table import MinimisationFailedError
+from rig.routing_table.remove_default_routes import minimise as \
+    remove_default_entries
+from rig.routing_table.ordered_covering import minimise as ordered_covering
+from six import iteritems
+
+
+def minimise(routing_tables, target_lengths,
```
mossblaser: Perhaps renaming this as "minimise_tables" would be better to distinguish it from the "minimise" functions of each algorithm(?)

rig/routing_table/ordered_covering.py
```diff
@@ -124,12 +124,15 @@
 """
 from collections import defaultdict, namedtuple
 from rig.routing_table import MinimisationFailedError, RoutingTableEntry
+from rig.routing_table.remove_default_routes import \
+    minimise as remove_default_routes
+from rig.routing_table.utils import intersect
 from six import itervalues
 
 
 def minimise(routing_table, target_length=1024):
```
mossblaser: Remove default argument value for target_length.

rig/routing_table/ordered_covering.py
```diff
@@ -329,7 +391,7 @@ def gg(entry):
 
 class Merge(namedtuple("Merge", ["routing_table", "entries", "key", "mask",
```
mossblaser: If everything else is being underscore-prefixed, this probably should be too.

rig/routing_table/utils.py
```diff
@@ -65,12 +98,57 @@ def table_is_subset_of(entries_a, entries_b):
...
+            # table never matched an entry in the second table. If the entry
+            # from the first table could not be default routed we return False
+            # as the tables cannot be equivalent.
+            default_routed = False
+
+            if len(entry.route) == 1 and len(entry.sources) == 1:
+                source = next(iter(entry.sources))
+                sink = next(iter(entry.route))
+
+                if source is not None and source is sink.opposite:
```
mossblaser: Should have an "is_link" check somewhere, no?

rig/place_and_route/utils.py
mundya: Done.

mundya:
@mossblaser I think this covers everything. After your doc change I'll squash before we merge back into the `routing-table-tools` branch.

mossblaser:
@mundya if you wouldn't mind running a light review on 114cde5 that would be great! Thanks! Documentation is on its way in a bit...

mossblaser:
The documentation should now have been updated and improved. @mundya some review would be welcome! I *think* I've caught all references to old names...

mossblaser:
OK, I believe this PR LGTM once the tests finally get around to passing... Since I believe this is a fast-forward for #211, that PR can also get merged into master.

PR #214, opened 2016-01-13
Fix flake8 issues

mossblaser:
These (somehow) have only started being flagged by a new version of flake8
which has seemingly only just started being used on travis.


mundya:
Cheers

PR #215, opened 2016-01-13
Fix oversight in ordered covering

mundya:
Bug would have meant that removing default entries would not have been attempted if ordered covering failed to produce a table at least as small as the target length.


mossblaser:
LGTM!

mundya:
Thanks!

PR #218, opened 2016-01-28
Add corresponding __ne__ methods to objects with __eq__

mossblaser:
These are required for Python 2.7 compatibility: In Python 3.x if __ne__ is
omitted, it defaults to the inversion of __eq__ if __eq__ defined. In Python
2.x __ne__ defaults to id(a) != id(b).

Thanks to @alan-stokes for spotting this bug!


mundya:
LGTM, the failing tests look like a connectivity issue.

PR #221, opened 2016-02-05
Add get_location_and_length to BitField

mossblaser:
This new method allows the direct querying of the size and location allocated
to fields in a bit field, without having to deconstruct a mask.

This should address the need expressed by @alan-stokes in SpiNNakerManchester/PACMAN#35 while being less wasteful than just computing these values from the mask.

### DIFF
--- a/rig/bitfield.py
+++ b/rig/bitfield.py
@@ -416,6 +416,42 @@ class BitField(object):
         """
         return self.fields.get_field(field, self.field_values).tags.copy()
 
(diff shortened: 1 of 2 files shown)


mundya:
LGTM, merge when Travis is happy...

mundya:
Hmm... distinct lack of coverage and PR testing there...

mossblaser:
Yes :/ tested them offline myself in the end...

PR #217, opened 2016-01-18
C-based simulated annealing algorithm

mossblaser:
This PR adds a (much faster) C based SA algorithm to Rig. Since including this in the default rig installation would cause end users to have a working C compiler and libffi installation, the C-related parts of this code are located in an external package: [rig_c_sa](https://github.com/project-rig/rig_c_sa). The C implementation is used when available, falling back on the old Python implementation.

* [x] Refactor SA algorithm to enable inclusion of custom kernels.
* [x] Implement SA kernel in [another package](https://github.com/project-rig/rig_c_sa)
* [x] Implement + test C SA Kernel for Rig as an "extra"
* [x] Fix travis...
* [x] ~~Discuss: Is the pip "extras" mechanism (used in the latest commit) the best way to enable this feature or should we just say "install rig_c_sa to enable this feature"?~~ The extras mechanism in pip is definately rather [half-baked](http://stackoverflow.com/questions/4796936/does-pip-handle-extras-requires-from-setuptools-distribute-based-sources#comment35838236_18879288)... Switching to "install this extra package for better performance!"
* [x] Discuss: Should warnings be made when using Python based placement?
* [x] Discuss: Where should the installation process be first advertised? On the README? In the P&R section? Just in the sa placer's docs (as in this commit)?

PR #222, opened 2016-02-16
Insert entries above blocks of similar entries

mundya:
Insert entries above blocks of entries of equivalent generality. This
makes that the Python implementation of OC matches (a) the C version for
SpiNNaker (b) that described in the HPSR submission.

For table ordering, see
https://github.com/project-rig/rig_routing_tables/blob/master/include/ordered_covering.h#L16-L57

For downcheck preferring to modify more significant bits in merges see
https://github.com/project-rig/rig_routing_tables/blob/master/include/ordered_covering.h#L285-L292

For merge application, see
https://github.com/project-rig/rig_routing_tables/blob/master/include/ordered_covering.h#L409-L485

### DIFF
--- a/rig/routing_table/ordered_covering.py
+++ b/rig/routing_table/ordered_covering.py
@@ -488,13 +488,13 @@ class _Merge(namedtuple("_Merge", ["routing_table", "entries", "key", "mask",
 
         # Iterate through the old table copying entries acrosss
         insert = 0
-        for remove, entry in enumerate(self.routing_table):
+        for i, entry in enumerate(self.routing_table):
             # If this is the insertion point then insert
-            if remove == self.insertion_index:
+            if i == self.insertion_index:
                 new_table[insert] = new_entry
                 insert += 1
 
-            if remove not in self.entries:
+            if i not in self.entries:
                 # If this entry isn't to be removed then copy it across to the
                 # new table.
                 new_table[insert] = entry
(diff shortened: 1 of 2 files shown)


mundya:
@mossblaser - the coverage seems to indicate a problem with stuff touched by #217 rather than by this PR.

rig/routing_table/ordered_covering.py
```diff
@@ -467,33 +472,45 @@ def apply(self, aliases):
...
-            offset += 1
-            km = (entry.key, entry.mask)
-
-            # Add to the alias list
-            our_aliases.update(aliases.pop(km, set([km])))
-
-        return routing_entries, aliases
+        # Iterate through the old table copying entries acrosss
+        insert = 0
+        for remove, entry in enumerate(self.routing_table):
```
mossblaser: I think I would prefer "remove" to be "i" since "remove" does not say "index in the (old) table" to me.

rig/routing_table/ordered_covering.py
```diff
@@ -738,7 +755,7 @@ def _refine_downcheck(merge, aliases, min_goodness):
 
                 # Add this settable mask and the required values to the
                 # settables list.
-                bits_and_vals.update((bit, bool(key & bit)) for bit in
+                bits_and_vals.update((bit, not bool(key & bit)) for bit in
```
mossblaser: The "bool" is superfluous here due to the "not", but up to you as you may prefer it as a stylistic choice.

mossblaser:
Superficial criticisms aside this looks good to me.

The coverage issue (as mentioned offline) appears to be a result of coveralls doing something a bit strange and I'll investigate at some other time. As such I'm happy to ignore it for this PR since we've both verified the coverage locally is actually still 100%.

mossblaser:
> I've retained the brackets in `not (key & bit)` because I feel it's
> clearer than `not key & bit`.

I would agree :).

LGTM! Merge at will once travis is happy.

... I must figure out what's going on with coveralls at some point :/

mundya:
@rowleya / @alan-stokes - this should improve the compression of the ucolumn tables.

PR #223, opened 2016-02-22
Fix multiply-overlapping SameChipConstraint

mossblaser:
As spotted by @neworderofjamie, if a pair of SameChipConstraints overlap by
more than one vertex, the apply_same_chip_constraints utility function would
crash. This commit fixes this issue and updates the tests to explicitly test
this behaviour.


mossblaser:
@neworderofjamie approves; awaiting travis...

mossblaser:
The coverage issue here is an ongoing issue with coveralls, coverage has not actually dropped and this issue will be fixed separately.

PR #225, opened 2016-02-25
Make geometry functions return ints, not Numpy look-alikes

mossblaser:
This was causing bugs in other software (spalloc_server) which attempts to
serialise output from these functions as JSON. Regardless, this unintended and
non-obvious bit of behaviour is now resolved.


mundya:
LGTM

PR #226, opened 2016-02-26
Make a_star return Routes, not Links.

mossblaser:
Previously, use of A* would result in routing trees with some routes described
with Links objects, not Routes. For most of human history this didn't cause
anything to blow-up but with newer more sophisticated routing table generation
everything blows up. Whoops.


mundya:
Whoops indeed. LGTM, thanks.

PR #227, opened 2016-03-02
Makes the SA algorithm callback test deterministic.

mossblaser:
Previously this test was not deterministic leaving a small probability that it
would result in a false negative in certain unlikely, but possible, situations.


mundya:
LGTM, merge when ready

PR #229, opened 2016-03-14
Add free commands

mundya:
- [x] Free SDRAM by pointer
 - [x] Free file-like views of SDRAM
 - [x] ~~Free routing table entries~~ we can already remove all routing table entries associated with an app_id (see `MachineController:clear_routing_table_entries`).

Beat you to the punch @neworderofjamie.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -2352,11 +2352,12 @@ class SpiNNakerLoadingError(Exception):
 
 
 def _if_not_closed(f):
-    """Run the method iff. the memory view hasn't been closed."""
+    """Run the method iff. the memory view hasn't been closed and the parent
+    object has not been freed."""
     @add_signature_to_docstring(f)
     @functools.wraps(f)
     def f_(self, *args, **kwargs):
-        if self.closed:
+        if self.closed or self._parent._freed:
             raise OSError
         return f(self, *args, **kwargs)
 
@@ -2585,6 +2586,18 @@ class SlicedMemoryIO(object):
             )
 
 
+def _if_not_freed(f):
+    """Run the method iff. the memory view hasn't been closed."""
+    @add_signature_to_docstring(f)
+    @functools.wraps(f)
+    def f_(self, *args, **kwargs):
+        if self._freed:
+            raise OSError
+        return f(self, *args, **kwargs)
+
+    return f_
+
+
 class MemoryIO(SlicedMemoryIO):
     """A file-like view into a subspace of the memory-space of a chip.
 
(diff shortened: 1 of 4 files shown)


neworderofjamie:
Hurrah!

rig/machine_control/machine_controller.py
```diff
@@ -2433,6 +2446,16 @@ def close(self):
             self.flush()
             self.closed = True
 
+    def free(self):
+        """Free the memory represented by the file-like."""
```
mossblaser: Would it be worth crashing out if you try and free a subregion? I can only imagine that resulting in bugs(?).

The alternative might be freeing once all subregions are freed but that seems like an invitation for complexity and memory leaks...

tests/machine_control/test_machine_controller.py
```diff
@@ -539,6 +539,46 @@ def test_app_stop_and_count(self, controller):
...
+    @pytest.mark.order_after("live_test_load_application")
+    def test_sdram_alloc_and_free(self, controller):
+        """Test subsequent allocation, freeing and reallocation of a large
+        block of memory.
+
+        This is an implicit test that FREE actually works.
+        """
+        with controller(x=0, y=0):
+            # Allocate 90MiB
+            ptr = controller.sdram_alloc(90*(1 << 20))
```
mossblaser: Does this not require an assertion to make sure the allocation actually occurred?

(I would probably also include a second alloc which you assert for failure, as a sanity check).

rig/machine_control/machine_controller.py
```diff
@@ -1103,6 +1103,18 @@ def sdram_alloc_as_filelike(self, size, tag=0, x=Required, y=Required,
...
+    @ContextMixin.use_contextual_arguments()
+    def sdram_free(self, ptr, x=Required, y=Required, app_id=Required):
+        """Free an allocated block of memory in SDRAM.
+
+        Parameters
+        ----------
+        ptr : int
+            Address of the block of memory to free.
+        """
+        arg1 = (app_id << 8) | consts.AllocOperations.free_sdram_by_ptr
```
mossblaser: What does SC&MP do with the app-id? Maybe omit it if nothing?

rig/machine_control/machine_controller.py
mundya: That's an excellent point, I'll remove the commit for this from the PR for the time being.

tests/machine_control/test_machine_controller.py
mundya: I think we do have some tests elsewhere that allocs actually work, but repeating them here as a sanity check is a good idea.

rig/machine_control/machine_controller.py
mundya: Oh yeah, thanks.

tests/machine_control/test_machine_controller.py
mundya: See below.

mossblaser:
Freeing stuff LGTM :). Following offline discussion:

* File-likes should probably be free-able, though only the 'root' file-like should actually be allowed to do the freeing.
* A check should be performed to ensure the memory associated with a file-like is not freed before all read/write operations.

mundya:
Note: this PR now also removes the buffering of file-like views of SDRAM.

mossblaser:
Minus failing tests, this looks rather sensible!

mundya:
@neworderofjamie - does this seem OK to you? Particularly the loss of buffering and the requirement that only the `MemoryIO` returned by `sdram_alloc_as_filelike` (and not any gained by slicing) can be freed.

neworderofjamie:
I think you convinced me of the need to remove it so I have resigned my self to the loss of buffering and that seems an entirely logical restriction on freeing memory allocated using ``sdram_alloc_as_filelike``

mossblaser:
Sounds like this is rigged then! Merge at will, @mundya.

mundya:
I've just squashed that down a bit, I'll merge when Travis is happy. Thanks @mossblaser

PR #230, opened 2016-03-14
Fix non-determinstic behaviour in CKernel initialisation

mossblaser:
Previously iterated over the 'fixed_vertices' set whose order is not fixed...

Should also fix an occasionally failing test...


mundya:
LGTM, merge at will...

PR #231, opened 2016-03-14
Deterministic __str__ for routing table entries

mundya:
Previously the source component of "X Y Z -> XX...X -> A B C" was
non-deterministic.


mundya:
Motivation: I had some failing tests on my laptop.

mossblaser:
LGTM, merge at will!

What test was dependent on this? A doctest?

mundya:
It was https://github.com/project-rig/rig/blob/master/tests/routing_table/test_entries.py#L47 specifically.

PR #232, opened 2016-03-15
Switch to codecov for coverage checking.

mossblaser:
Previously used coveralls which has been increasingly unreliable.

OK, so I think this change is 100% positive:

* Codecov, like coveralls, is free for open source projects.
* Though it also mistakenly reports coverage on omitted files, it is trivial to inform the webpage that a file should be omitted.
* It actually understands branch coverage(!)
* It actually understands the distinction between branch and PR builds.
* It actually understands that travis runs multiple builds of each branch/PR
* It actually discovers, without hints, where the source files are (even though the coverage file reports them as being somewhere funny)
* Generally much nicer web UI
* Much better documentation

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -7,7 +7,7 @@ python:
 install:
         - python setup.py install
         - pip install -r requirements-test.txt
-        - pip install python-coveralls
+        - pip install codecov
         - pip install spinnaker_proxy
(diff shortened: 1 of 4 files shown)


mundya:
Rigged

Awesome, merge when ready!

PR #233, opened 2016-03-15
Make discover_connections more fault-tolerant.

### LINKED ISSUE #228: Make `discover_connections` more fault tolerant.
If an ethernet-connected chip dies (or has its P2P route broken) before `discover_connections()` runs, getting the chip's IP will cause the whole function to die.

mossblaser:
Fixes #228.


mundya:
LGTM

Issue #131, opened 2015-05-01
Prune dead code


mundya:
- [x] We don't use `spinX.conf` files so remove:
   - https://github.com/project-rig/rig/blob/master/rig/machine_control/struct_file.py#L92-L125
   - https://github.com/project-rig/rig/blob/master/tests/machine_control/test_struct_file.py#L169-L201
   - _Worth noting that these files will be touched by #82, so it might be worth deprecating this method and leaving pruning it until #82._

PR #235, opened 2016-03-17
Port to SpiNNaker Tools v2.0.0

mossblaser:
**I propose that once merged this marks v1.0.0 of Rig which we really should have reached ages ago...**

* [x] Code review.
* [x] Plonk in the official released files from SpiNNaker Tools v2.0.0.

This commit migrates over to a release-candidate of SpiNNaker Tools 2.0.0 which
includes and new SC&MP which notably no longer needs to know the size of the
machine when booting nor requires that chip (0, 0) exist at all.

**NOTE** This commit includes an unofficial SpiNNaker Tools release candidate
version of SC&MP, sark.struct and test application binary. These should be
rebuilt using the officially released SpiNNaker Tools package before final
release of these changes to Rig.

Changes:

* Boot-related changes:
  - MachineController.boot() may be called without arguments (giving a width
    and height result in a deprecation warning and the values being ignored.
  - The boot() function, by default, waits for the machine to report that it is
    completely booted before returning.
  - The boot functions now accept filenames for the boot and struct files.
  - Additional kwargs to boot functions now override values in the SV struct.
  - As a result, the custom arguments for setting the hardware version and LEDs
    (etc.) have been removed in favour of directly specifying the SV values. I
    am not aware of these actually being used anywhere.
  - The rig-boot command no longer takes any arguments indicating the size of
    the machine. Calling it with a width or height is now an error.
  - Associated documentation has been updated.

* Other machine-control changes:
  - Both new- and old-format version formats is understood by
    get_software_version for SpiNNaker and BMPs alike.
  - Version numbers are now reported as tuples (major, minor, patch) rather
    than floats. (A breaking change)
  - System-wide commands (e.g. signalling etc.) are now sent to chip (255,
    255), an alias for the 'root chip' used to boot the machine rather than (0,
    0) which may not exist.
  - A new 'root_chip' attribute will report the (x, y) coordinate of the chip
    used to boot the machine. This is useful for certain geometry functions
    (e.g. discovery of Ethernet chip locations).

* All SpiNN-5 board related geometry functions now take the coordinates of the
  root chip as arguments. This is important when doing calculations relating to
  machines which do not have a chip (0, 0). All internal uses of these
  functions have been updated to provide this information. The one example use
  of these functions in the documentation has been updated.

* Hints about the once-missing sark_tag_ptr function have been removed with the
  expectation that new applications will use Spinnaker Tools v2.0.0 which now
  includes this function.

* The test binary has been rebuilt using SpiNNaker Tools v2.0.0 and has a new
  Makefile for building against this version of the tools.

* The test suite now does not accept a width and height when running against
  real SpiNNaker hardware. The developer documentation and travis scripts have
  been updated accordingly.


mundya:
> MachineController.boot() may be called without arguments (giving a width and height result in a deprecation warning and the values being ignored.

Given that this is a major version bump I'd be happy skipping the deprecation stage and just remove those arguments.

Other than that this all sounds awesome and I'll review!

mossblaser:
> Given that this is a major version bump I'd be happy skipping the deprecation stage and just remove those arguments.

I'm leaving them since a good number of rig scripts are floating around which I would wish to remain working. Plus it means that doing pip upgrade will allow nengo_spinnaker/Pinus-Rigida to benefit from hardware black-listing, faster booting, and general SC&MP robustness improvements the second this is released without modification.

mundya:
> I'm leaving them since a good number of rig scripts are floating around which I would wish to remain working. Plus it means that doing pip upgrade will allow nengo_spinnaker/Pinus-Rigida to benefit from hardware black-listing, faster booting, and general SC&MP robustness improvements the second this is released without modification.

Fair enough, thanks.

mossblaser:
NB: this fixes #131. I suggest de-milestoning the other 1.0 milestones since they are not blocking and all have potential implementations with no breaking changes.

rig/machine_control/machine_controller.py
```diff
@@ -276,10 +296,14 @@ def boot(self, width, height,
...
                 supplied boot image, even if ``only_if_needed`` is False.
-
         check_booted : bool
-            If ``check_booted`` is True this method checks if the machine was
-            successfully booted. If False, this check is skipped.
+            If ``check_booted`` is True this method waits for the machine to be
+            fully booted before returning. If False, this check is skipped and
+            the function returns as soon as the machine's Ethernet interface is
+            likely to be up (but not necessarily before booting has completed).
+        **sv_defaults : {name: value, ...}
```
mundya: `defaults` seems like a bad name for something that overrides defaults?

rig/machine_control/machine_controller.py
```diff
@@ -326,14 +355,17 @@ def boot(self, width, height,
...
-        # Check the machine is now booted
+        # Wait for the machine to completely boot
         if check_booted:
             try:
-                self.get_software_version(0, 0, 0)
+                p2p_address = (255, 255)
+                while p2p_address == (255, 255):
+                    time.sleep(0.1)
+                    p2p_address = self.get_software_version(
+                        255, 255, 0).position
```
mundya: Could you not store the value of `_root_chip` here?

rig/machine_control/machine_controller.py
```diff
@@ -2139,10 +2179,10 @@ def __contains__(self, chip_core_or_link):
 
 class ProcessorStatus(collections.namedtuple(
     "ProcessorStatus", "registers program_state_register stack_pointer "
-                       "link_register rt_code cpu_flags cpu_state "
+                       "link_register rt_code phys_cpu cpu_state "
                        "mbox_ap_msg mbox_mp_msg mbox_ap_cmd mbox_mp_cmd "
                        "sw_count sw_file sw_line time app_name iobuf_address "
-                       "app_id user_vars")):
+                       "app_id version user_vars")):
```
mundya: `version` -> `software_version`?

rig/wizard.py
```diff
@@ -102,6 +91,12 @@ def dimensions_wizard():
     """A wizard which attempts to determine the dimensions of a SpiNNaker
     system.
 
+    .. warning::
+
+        Since SC&MP v2.0.0, it is not necessary to know the dimensions of a
```
mundya: ~~","~~  ... being very picky now

mundya:
I'm very pleased at how much this change simplifies things! This pretty much LGTM!

mundya:
> I propose that once merged this marks v1.0.0 of Rig which we really should have reached ages ago...

> I suggest de-milestoning the other 1.0 milestones since they are not blocking and all have potential implementations with no breaking changes.

Agreed to both.

rig/machine_control/machine_controller.py
mossblaser: In fact, I could set a whole load of similar variables here...

mossblaser: On balance, how strongly would you feel about this not being added? It should not be noticeable in terms of performance for anything and the existing code path needs to remain in existence for when you're not booting anyway so it seems like there's no benefit to complicating matters?

mundya: That's fine by me; I'll just re-grok the existing code path (didn't spot where `_root_chip` was set).

rig/machine_control/machine_controller.py
mossblaser: Done.

rig/machine_control/machine_controller.py
mossblaser: Should be set by the property when it is None. (The function which decides what SCP connection to use only uses _root_chip if it has been set since sending the command would require looking up the SCP connection to use which would [....])

mundya:
2f0c20f39561dd7c795d02c8b206a66318653167 looks fine to me.

PR #237, opened 2016-03-23
Memory io improvements

mossblaser:
`MemoryIO` reads/writes which are truncated now produce a warning. These can, if desired, be turned into exceptions by an application.

### DIFF
--- a/rig/machine_control/machine_controller.py
+++ b/rig/machine_control/machine_controller.py
@@ -2584,10 +2584,9 @@ class SlicedMemoryIO(object):
         if self.address + len(bytes) > self._end_address:
             n_bytes = self._end_address - self.address
 
-            if len(bytes) > n_bytes:
-                warnings.warn("write truncated from {} to {} bytes".format(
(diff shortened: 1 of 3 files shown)


neworderofjamie:
Looks entirely sensible

rig/machine_control/machine_controller.py
```diff
@@ -2568,13 +2582,15 @@ def write(self, bytes):
...
         """
         if self.address + len(bytes) > self._end_address:
-            n_bytes = min(len(bytes), self._end_address - self.address)
-
-            if n_bytes <= 0:
-                return 0
+            n_bytes = self._end_address - self.address
 
+            warnings.warn("write truncated from {} to {} bytes".format(
+                len(bytes), n_bytes), TruncationWarning)
```
mundya: Setting the stack level here would be useful, should I open a PR?

mossblaser: Please do :)

PR #238, opened 2016-03-29
Fix reference to CoreInfo.version in rig_info

mundya:
Should have instead referenced `CoreInfo.software_version`.


tests/scripts/test_rig_info.py
```diff
@@ -39,7 +40,7 @@ def test_unknown_arch(monkeypatch):
     mc = mock.Mock()
     info = mock.Mock()
```
mundya: Strictly I should make this `info = mock.Mock(spec_set=CoreInfo)` too?

mossblaser: Presumably... I've been pretty lax about `spec_set`ing things... :(

mundya:
Thanks :smiley:

PR #239, opened 2016-03-30
Warn when saturating values

mundya:
Modifies the typecast methods so that they warn when a value has been saturated. This is mostly intended as being a useful heads-up for developers rather than something most users will encounter.

Thoughts @mossblaser, @neworderofjamie?


mossblaser:
Seems legit to me, though I don't often use this stuff! :)

mundya:
Closing this on the basis that I should do something better.

PR #240, opened 2016-04-05
Having a fixed-point format that can't represent a full range of valu…

neworderofjamie:
I think it's entirely legitimate to define a fixed point format where the position of the fixed point is not within your representable range small weights


mundya:
Sure, can you fix the tests?

mundya:
On reflection this may be a good opportunity to fix some of the annoyances of the way float_to_fix and fix_to_float are implemented. I'm playing with that now.

PR #242, opened 2016-04-14
Add get_iobuf_bytes.

mossblaser:
Allows access of raw IOBUF data, regardless of encoding.

I was trying to think of a sensible way to make the default "crash" in the presence of bad unicode data and couldn't think of anything. Instead as a stop-gap (and because it strictly was missing) I've just added an additional iobuf-getting function which returns raw bytes...


mundya:
LGTM!

PR #245, opened 2016-05-13
Improve type cast utilities

mundya:
New methods `float_to_fp` and `fp_to_float` work with signed and unsigned values as expected. New methods and `NumpyFloatToFixConverter` can now have negative numbers of fractional bits or more fractional bits than can be represented in the result - as requested by @neworderofjamie.

### DIFF
--- a/rig/type_casts.py
+++ b/rig/type_casts.py
@@ -1,11 +1,118 @@
 """Fixed point conversion utilities.
 """
 import numpy as np
+import warnings
+
+
+def float_to_fp(signed, n_bits, n_frac):
+    """Return a function to convert a floating point value to a fixed point
+    value.
+
+    For example, a function to convert a float to a signed fractional
+    representation with 8 bits overall and 4 fractional bits (S3.4) can be
+    constructed and used with::
+
+        >>> s34 = float_to_fp(signed=True, n_bits=8, n_frac=4)
+        >>> hex(int(s34(0.5)))
+        '0x8'
+
+    The fixed point conversion is saturating::
+
+        >>> q34 = float_to_fp(False, 8, 4)  # Unsigned 4.4
+        >>> hex(int(q34(-0.5)))
+        '0x0'
+
+        >>> hex(int(q34(15.0)))
+        '0xf0'
+
+        >>> hex(int(q34(16.0)))
+        '0xff'
+
+    Parameters
+    ----------
+    signed : bool
+        Whether the values that are to be converted should be signed, or
+        clipped at zero.
+
+            >>> hex(int(float_to_fp(True, 8, 4)(-0.5)))  # Signed
+            '-0x8'
+            >>> hex(int(float_to_fp(False, 8, 4)(-0.5)))  # Unsigned
+            '0x0'
+
+    n_bits : int
+        Total number of bits in the fixed-point representation (including sign
+        bit and fractional bits).
+    n_frac : int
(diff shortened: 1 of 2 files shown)


mundya:
Should replace #240.

neworderofjamie:
Thank god, I had just beaten pytest into submission and doctest had raised
it's magical head...

neworderofjamie:
This looks way better to me! Will merge once the tests pass n'shit then can you make a release @mossblaser ? It would be nice to make a pynn_spinnaker release depending on a version of rig which won't complain in (to the user) mysterious ways

mundya:
@neworderofjamie - I think I'll need to fix a doctest, and squash, but I'll let you know when I've done that. Sorry for being zlow.

mossblaser:
Sounds good! I'll release once this is all reviewed and merged!

rig/type_casts.py
```diff
@@ -1,16 +1,123 @@
...
+    def bitsk(value):
+        """Convert a floating point value to a fixed point value.
+
+        Parameters
+        ----------
+        value : float
+            The value to convert.
+        """
+        int_val = int(scale * value)
+        return np.clip(int_val, min_v, max_v)
```
mossblaser: Does np.clip return real Python integers or numpy look-alikes? (I've got into trouble with this in the geometry module before...)

rig/type_casts.py
```diff
@@ -1,16 +1,123 @@
...
+    value.
+
+    For example, a function to convert from signed fractional representations
+    with 4 fractional bits constructed and used with::
+
+        >>> f = fp_to_float(4)
+        >>> f(0x08)
+        0.5
+
+        >>> f(-0x8)
```
mossblaser: I presume the assumption here is that values are unpacked as signed numbers too and get the sign put on at that stage?

rig/type_casts.py
neworderofjamie: I had this precise fear, on my machine it seems to work correctly, but on numpy 0.8 who knows

rig/type_casts.py
```diff
@@ -1,16 +1,123 @@
...
+
+    return kbits
+
+
+def float_to_fix(signed, n_bits, n_frac):
+    """**DEPRECATED** Return a function to convert a floating point value to a
+    fixed point value.
+
+    .. warning::
+        This function is deprecated in favour of :py:meth:`~.float_to_fp`.
```
mossblaser: Could you give a hint here as to *what* has changed to lead to this deprecation? Scanning the documentation very quickly I couldn't work it out, sorry!

mundya: The replacement functions are closer to what you'd expect :/

rig/type_casts.py
mundya: Yeah, it's like the Numpy equivalent below: if you pass it something signed then it'll respect that, otherwise it'll treat it as unsigned.

And yeah, `struct.unpack("i", ...)` is what I'm expected to be used.

rig/type_casts.py
mundya: Good point, I'll change that

rig/type_casts.py
mossblaser: Can that be defined (or an example given?)

neworderofjamie:
Tested against pynn_spinnaker, works great

PR #246, opened 2016-05-13
Rename codecov.yaml to codecov.yml

mundya:


mossblaser:
LGTM and no shouting from codecov to be seen! Merge at willl...

PR #248, opened 2016-05-18
Add support for Semantic Version label postfixes.

mossblaser:
Adds a new field to the CoreInfo and BMPInfo namedtuples which contains any
version postfixes (e.g. -dev or build numbers).

Also refactors version-number parsing so that the implementation is shared
between BMPController and MachineController.

### DIFF
--- a/rig/machine_control/bmp_controller.py
+++ b/rig/machine_control/bmp_controller.py
@@ -190,7 +190,7 @@ class BMPController(ContextMixin):
         can_id = (sver.arg1 >> 8) & 0xff
         board_id = sver.arg1 & 0xff
 
-        # arg2 (version field unpacked seperately)
+        # arg2 (version field unpacked separately)
         buffer_size = (sver.arg2 & 0xffff)
 
         software_name, version, version_labels = \
@@ -408,7 +408,7 @@ class BMPInfo(collections.namedtuple(
         The position of the board in a frame. (This should correspond exactly
         with a board's board-coordinate.
     version : (major, minor, patch)
-        Software version number. See also: ``software_version_labels``.
+        Software version number. See also: ``version_labels``.
     buffer_size : int
         Maximum supported size (in bytes) of the data portion of an SCP packet.
     build_date : int
--- a/rig/machine_control/bmp_controller.py
+++ b/rig/machine_control/bmp_controller.py
@@ -4,13 +4,14 @@ import time
 import struct
 import collections
 
-from .scp_connection import SCPConnection
+from rig.machine_control.scp_connection import SCPConnection
 
-from . import consts
+from rig.machine_control import consts
 
-from .consts import SCPCommands, LEDAction, BMPInfoType, BMP_V_SCALE_2_5, \
-    BMP_V_SCALE_3_3, BMP_V_SCALE_12, BMP_TEMP_SCALE, BMP_MISSING_TEMP, \
-    BMP_MISSING_FAN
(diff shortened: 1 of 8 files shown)


mossblaser:
What on earth is going on with the tests :s can't reproduce locally atm... :s

mossblaser:
Ah... helps if you add new files... /me is an idiot as usual...

mossblaser:
(And this is why continuous integration is good :P)

rig/machine_control/bmp_controller.py
```diff
@@ -416,7 +408,7 @@ class BMPInfo(collections.namedtuple(
         The position of the board in a frame. (This should correspond exactly
         with a board's board-coordinate.
     version : (major, minor, patch)
-        Software version number.
+        Software version number. See also: ``software_version_labels``.
```
mundya: Where is `software_version_labels`? Don't need the `software`?

rig/machine_control/machine_controller.py
```diff
@@ -474,23 +475,14 @@ def get_software_version(self, x=255, y=255, processor=0):
         pcpu = (sver.arg1 >> 8) & 0xff
         vcpu = sver.arg1 & 0xff
 
-        # arg2 => version number and buffer size
-        version = (sver.arg2 >> 16)
+        # arg2 => version number (parsed seperately) and buffer size
```
mundya: sp. separately

rig/machine_control/bmp_controller.py
mossblaser: Whoops... Spot the copy-paste!

rig/machine_control/common.py
```diff
@@ -0,0 +1,62 @@
+"""Internal-use functions shared between the
+:py:class:`~rig.machine_control.MachineController` and
+:py:class:`~rig.machine_control.BMPController` objects.
+"""
+
+import re
+
+
+VERSION_NUMBER_REGEX = re.compile(r"^(\d+)[.](\d+)[.](\d+)(\D.*)?$")
```
mundya: Is it not fair to assume that the tags will begin with a "-", that is the spec I think?

mossblaser: Not quite. For example "1.2.3+9fb2b01" is valid, as is "1.2.3-dev+9fb2b01"

rig/machine_control/common.py
```diff
@@ -0,0 +1,62 @@
...
+        minor = legacy_version_field % 100
+        patch = 0
+        labels = ""
+    else:
+        # Semantic Version encoding: packed after the null-terminator of the
+        # software name in the version string.
+        software_name, _, version_number = software_name.partition("\0")
+
+        match = VERSION_NUMBER_REGEX.match(version_number.rstrip("\0"))
+        assert match, "Malformed version number: {}".format(version_number)
```
mundya: Is it worth testing that this `assert` is present/replacing it with a more formal error?

rig/machine_control/common.py
mundya: Cool, thanks

rig/machine_control/common.py
mossblaser: I'm split on this one. Usually yes but in this case a malformed version probably indicates something really bad is going on so an assert suits fine. The assert is only there to make the explosion more useful than "None has no .group()".

mundya:
Looks good to me apart from the copy-paste and the spelling of "separately".

PR #249, opened 2016-05-24
Skip alias check during default-route removal if possible.

mossblaser:
This optimisation makes table generation significantly faster by skipping O(N^2) alias checks during default-route removal when it can be cheaply proven that no aliases exist. In this commit, this is only possible when an input table contains entries with the same mask and no repeated keys.

In some experiments I am doing this brings routing table generation down from over a minute and a half to around 10 seconds.


mundya:
Rigged

Feel free to merge when tests are happy!

PR #253, opened 2016-06-09
Fix accidental-broadcast in circuit simulator example.

mossblaser:
A classic Python beginner's error: defining a function with a default argument
set to an empty list. As a result, every Wire created is connected to *every*
input in the circuit. How embarrassing.

For small circuits this bug doesn't break anything  since gates ignore packets
with unrecognised keys (hence not being spotted earlier) but for larger
circuits you get dreadful network contention...


mundya:
RIgged

PR #254, opened 2016-06-10
Add Reverse Cuthill-McKee (RCM) placement algorithm

mossblaser:
This simple breadth-first-search-like sequential placer can be found in the software placement literature and is believed to perform a reasonable job of placing pipeline-like things.

Sorry to drop you in reviewing this clearly thesis-specific change @mundya ...

### DIFF
--- a/rig/place_and_route/place/rcm.py
+++ b/rig/place_and_route/place/rcm.py
@@ -1,0 +1,171 @@
+"""Reverse Cuthill-McKee based placement.
+"""
+
+from collections import defaultdict
+
+from six import itervalues
+
+from rig.place_and_route.place.sequential import place as sequential_place
+from rig.links import Links
+from rig.netlist import Net
+
+
+def _get_vertices_neighbours(nets):
+    """Generate a listing of each vertex's immedate neighbours in an undirected
+    interpretation of a graph.
+
+    Returns
+    -------
+    {vertex: {vertex: weight, ...}), ...}
+    """
+    zero_fn = (lambda: 0)
+    vertices_neighbours = defaultdict(lambda: defaultdict(zero_fn))
+    for net in nets:
+        if net.weight != 0:
+            for sink in net.sinks:
+                vertices_neighbours[net.source][sink] += net.weight
+                vertices_neighbours[sink][net.source] += net.weight
+    return vertices_neighbours
+
+
+def _dfs(vertex, vertices_neighbours, _visited=None):
+    """Generate all the vertices connected to the supplied vertex in
+    depth-first-search order.
+    """
+    if _visited is None:
+        _visited = set()
+
+    yield vertex
+    _visited.add(vertex)
+
+    for neighbour in vertices_neighbours[vertex]:
+        if neighbour not in _visited:
+            for vertex in _dfs(neighbour, vertices_neighbours, _visited):
+                yield vertex
+
+
+def _get_connected_subgraphs(vertices, vertices_neighbours):
+    """Break a graph containing unconnected subgraphs into a list of connected
+    subgraphs.
+
+    Returns
+    -------
+    [set([vertex, ...]), ...]
+    """
+    remaining_vertices = set(vertices)
+    subgraphs = []
+    while remaining_vertices:
+        subgraph = set(_dfs(remaining_vertices.pop(), vertices_neighbours))
+        remaining_vertices.difference_update(subgraph)
+        subgraphs.append(subgraph)
+
+    return subgraphs
+
+
+def _cuthill_mckee(vertices, vertices_neighbours):
+    """Yield the Cuthill-McKee order for a connected, undirected graph.
+
+    `Wikipedia
+    <https://en.wikipedia.org/wiki/Cuthill%E2%80%93McKee_algorithm>`_ provides
+    a good introduction to the Cuthill-McKee algorithm. The RCM algorithm
+    attempts to order vertices in a graph such that their adjacency matrix's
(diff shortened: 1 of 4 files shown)


rig/place_and_route/place/rcm.py
```diff
@@ -0,0 +1,178 @@
...
+    for subgraph_vertices in _get_connected_subgraphs(vertices_resources,
+                                                      vertices_neighbours):
+        cm_order = _cuthill_mckee(subgraph_vertices, vertices_neighbours)
+        for vertex in reversed(cm_order):
+            yield vertex
+
+
+def rcm_chip_order(machine):
+    """A generator which iterates over a set of chips in a machine in
+    Reverse-Cuthil-McKee order.
```
mundya: sp. Cuthill

rig/place_and_route/place/rcm.py
```diff
@@ -0,0 +1,178 @@
...
+    vertices_neighbours = _get_vertices_neighbours(nets)
+    for subgraph_vertices in _get_connected_subgraphs(vertices_resources,
+                                                      vertices_neighbours):
+        cm_order = _cuthill_mckee(subgraph_vertices, vertices_neighbours)
+        for vertex in reversed(cm_order):
+            yield vertex
+
+
+def rcm_chip_order(machine):
+    """A generator which iterates over a set of chips in a machine in
```
mundya: The thing this returns yields vertices, not chips - or am I misreading?

tests/place_and_route/place/test_rcm.py
```diff
@@ -0,0 +1,219 @@
...
+
+    # Seperate pairs of vertices
+    v0 = "v0"
+    v1 = "v1"
+    v2 = "v2"
+    v3 = "v3"
+    vertices_resources = {v0: {}, v1: {}, v2: {}, v3: {}}
+    nets = [Net(v0, v1), Net(v2, v3)]
+    assert list(rcm_vertex_order(vertices_resources, nets)) in (
+        # v1 & v2 first, all possible orders of each pair
```
mundya: 0 & 1?

tests/place_and_route/place/test_rcm.py
```diff
@@ -0,0 +1,219 @@
...
+        [v0, v1, v2, v3],
+        [v0, v1, v3, v2],
+        [v1, v0, v2, v3],
+        [v1, v0, v3, v2],
+        # v3 & v4 first, all possible orders of each pair
```
mundya: 2 & 3?

mundya:
No test for the placer yet?

rig/place_and_route/place/rcm.py
mossblaser: You may be being misled: the implementation internally calls `rcm_vertex_order` to generate the ordering however it generates a graph in which the vertices are `(x, y)` chip coordinates and the nets represent working connections between those chips. As a result, the ordering produced by `rcm_vertex_order` is actually an ordering of chip coordinates.

Does that make sense? (What would be a better way of making this clearer in the code?)

tests/place_and_route/place/test_rcm.py
mossblaser: Thank you! Whoops!

mundya:
Looks good to me (not sure I fully understand) apart from the comments above.

mossblaser:
> No test for the placer yet?

There is a generic test suite for placement algorithms to which this placer has been added: https://github.com/project-rig/rig/pull/254/files#diff-e60413cd5d0cfb7adc8ce7a7f61c02e7R63 This suite should be sufficient, especially since this placer is really just a wrapper around the seqential placer in which the order in which vertices/chips are assigned is determined (in this placer) by the RCM algorthm.

rig/place_and_route/place/rcm.py
```diff
@@ -0,0 +1,182 @@
...
+    order.
+
+    For use as a vertex ordering for the sequential placer.
+    """
+    vertices_neighbours = _get_vertices_neighbours(nets)
+    for subgraph_vertices in _get_connected_subgraphs(vertices_resources,
+                                                      vertices_neighbours):
+        cm_order = _cuthill_mckee(subgraph_vertices, vertices_neighbours)
+        for vertex in reversed(cm_order):
+            yield vertex
```
mundya: If `vertex` is a chip... then maybe it could be called `chip`?

mossblaser: The `rcm_vertex_order` function's primary role is in fact for iterating over vertices in an application's graph. That said, as noted previously it is, in addition, internally used by `rcm_chip_order`.

Background: the `sequential` placer iterates over the vertices in an application graph and assigns them to chips in a machine, filling up one chip before moving on to the next. The order in which it iterates over vertices and fills chips is something the caller can control by providing an iterator over the vertices and chips respectively. The `breadth_first` placer simply provides a breadth-first-search-based iterator over vertices but doesn't specify a particular order in which to fill chips. The `hilbert` placer uses the same breadth-first vertex iterator used by the `breadth_first` placer but also provides an iterator over chips in a machine which causes the sequential placer to fill chips in hilbert-order.

The `rcm` placer simply defines that vertices should be placed in RCM order additionally that chips should be filled in RCM order. As such it implements two iterators `rcm_vertex_order` and `rcm_chip_order` which iterate over vertices in an application and chips in a machine (respectively) in RCM order. These iterators are then passed to the sequential placer which does the rest.

mundya:
Cool, I think I understand this now, thank you!

mundya:
Rigged

PR #255, opened 2016-06-10
Fix sequential placer SameChipConstraint edge-cases.

mossblaser:
Prior to this fix, if a placement problem contains overlapping or duplicate-containing `SameChipConstraints`, the `sequential` placer (and thus all other placers built on it, e.g. `hilbert` and `breadth_first`) would fall over.

Though these smell of 'edge-case' they are actually very common in PyNN SpiNNaker's netlists...  The generic placer test suite has been updated with a test case which include these behaviours..


mundya:
LGTM!

PR #256, opened 2016-06-10
Faster SCP

mundya:
Some tweaks to SCP performance
 - Use the packet transmission latency to hide the callback cost
 - Remove generators from read/write fast-loops [I'm actually quite sad about this]

Changes lead to ~5% performance improvement.


mossblaser:
Sigh! These changes look reasonable to me, though as disappointing as you'd expect, however... Feel free to merge when travis is happy...

PR #257, opened 2016-06-13
Fix stack-overflow in RCM placer for large netlists.

mossblaser:
Switches an internal depth-first-search function to use a deque to maintain the
search stack rather than using the call stack. This prevents stack overflows in
networks for which a depth-first search can visit extremely large numbers of
nodes.


mundya:
LGTM! Merge away...

PR #258, opened 2016-06-23
Upgrade to SC&MP 3.0.0

mossblaser:
* [x] Add SC&MP 3.0.0 boot image
* [x] Rebuild Rig test program
* [x] Add support for additional info included in ChipInfo in v3.0.0 of SC&MP.
* [x] Iterate over Ethernet connected chips using `SystemInfo`

This PR will unfortunately result in the release of Rig v2.0.0 since a version-checking bug in SC&MP 2.x.x series made binaries built against older versions erroneously incompatible with newer releases of SC&MP. No breaking changes in Rig are intended...


rig/geometry.py
```diff
@@ -323,6 +337,25 @@ def spinn5_local_eth_coord(x, y, w, h, root_x=0, root_y=0):
...
+            >> from rig.machine_control import MachineController
+            >> mc = MachineController("my-machine")
+            >> si = mc.get_system_info()
+            >> print(si[(3, 2)].local_ethernet_chip)
+            (0, 0)
+
+        :py:func:`.spinn5_local_eth_coord` will always produce the coordinates
+        of the Ethernet-connected SpiNNaker chip on the same SpiNN-5 board as
+        the supplied chip. In future versions of the low-level system software,
+        some other method of chosing local Ethernet connected chips may be
```
mundya: "chosing" -> "choosing"

rig/machine_control/machine_controller.py
```diff
@@ -2041,6 +2049,17 @@ class ChipInfo(collections.namedtuple(
...
+        The IP address of the Chip's Ethernet connection. If ethernet_up is
+        False, the value of this field is unpredictable and should be ignored.
+    local_ethernet_chip : (x, y)
+        The coordinates of the 'nearest' Ethernet connected chip to this chip,
+        corresponding with the value in ``sv->eth_addr``.
+
+        .. note::
+
+            This value may not literally be the *nearest* Ethernet connected
+            chip.
```
mundya: ...but instead may be?

rig/machine_control/machine_controller.py
```diff
@@ -2110,6 +2132,19 @@ def chips(self):
...
+    def ethernet_connected_chips(self):
+        """Iterate over the coordinates of Ethernet connected chips.
+
+        Yields
+        ------
+        ((x, y), str)
+            The coordinate and IP address of each Ethernet connected chip in
+            the system.
+        """
+        return ((xy, chip_info.ip_address)
```
mundya: I'm not really a fan of returning the generator, sorry.

I think

```python
for xy, chip_info in six.iteritems(self):
    if chip_info.ethernet_up:
       yield xy, chip_info
```

would be clearer

mossblaser: Presumably you mean yielding `(xy, chip_info.ip_address)` rather than changing this to `(xy, ChipInfo)`? I had considered the latter but couldn't think of a good use case.

mundya: Yeah, sorry, I did mean `chip_info.ip_address` - I mostly meant that I felt explicit use of the `yield` keyword would be better.

mossblaser:
Thanks @mundya!

mundya:
!(Rigged)[http://jhnet.com/misc/rigged.png] ... though it's a shame about the version number!

PR #260, opened 2016-07-18
Fix non-deterministic behaviour in SA algorithm.

mossblaser:
This bugfix makes the SA algorithm's output reproducible if given the same
graph and a random number generator with the same seed.

### DIFF
--- a/tests/place_and_route/place/sa/test_algorithm.py
+++ b/tests/place_and_route/place/sa/test_algorithm.py
@@ -4,13 +4,19 @@ from mock import Mock
 
 from collections import OrderedDict
 
-from six import itervalues
+from six import itervalues, iteritems
 
 import random
 
 from rig.place_and_route import Machine, Cores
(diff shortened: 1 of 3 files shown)


tests/place_and_route/place/sa/test_algorithm.py
```diff
@@ -62,6 +68,35 @@ def fn(iteration_count, placements, current_cost, r_accept,
         assert len(cb.mock_calls) > 1
 
 
+@pytest.mark.parametrize("kernel", [CKernel, PythonKernel])
+def test_deterministic(kernel):
+    """Ensure that placement is deterministic when a single seed is used."""
+    # Generate a random placement in which the graph is always the saame but
```
mundya: "saame

mundya:
LGTM, thank you

mossblaser:
Thanks very much @mundya.

PR #262, opened 2016-07-19
Scale HPWL by sqrt(n) in SA placer.

mossblaser:
This change makes the cost estimation function used by the SA placer more
accurate resulting in marginally better placements (see
project-rig/rig_c_sa#4).


mundya:
LGTM

PR #264, opened 2016-10-14
Fix ordering of flood-fill regions

mundya:
This is a bug that only appears with flood fills of large machines,
which is why it has only just shown up. The problem occurs because a
RegionTree iterates over its subregions recursively, whereas the entire
tree should be iterated over in increasing y before x.

For example, the selected regions:

	.... .... .... ....
	.... .... .... ....
	.... .... .... ....
	2... .... .... ....

	.... .... .... ....
	.... .... .... ....
	.... .... .... ....
	13.. .... .... ....

Should be produced in the order (1, 2, 3) but are actually produced in
the order (1, 3, 2) causing SC&MP to ignore region transmitted after 2.

This commit fixes the problem by including an additional call to
`sorted` in `compress_flood_fill_regions`. Ideally some better iteration
over the tree could be implemented. This needs further thought so I
suggest this fix for now.

Thanks to @lplana for bringing this bug to my attention.

### DIFF
--- a/rig/machine_control/regions.py
+++ b/rig/machine_control/regions.py
@@ -80,7 +80,7 @@ def compress_flood_fill_regions(targets):
         for p in cores:
             t.add_core(x, y, p)
 
-    return t.get_regions_and_coremasks()
+    return sorted(t.get_regions_and_coremasks())
 
 
 class RegionCoreTree(object):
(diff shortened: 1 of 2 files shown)


neworderofjamie:
Sorting the things which need to be sorted looks entirely sensible to me!

PR #265, opened 2016-10-26
Fix bug in machine boot kwargs

mundya:
Some of the keyword arguments used to boot SpiNNaker machines were overloaded. In particular, `boot_delay` is present both in the `sv` struct and in the list of arguments expected by `boot`. This commit fixes this issue by making it possible to set the SV overrides explicitly.

Note that the API changes are backwards compatible, this only adds to the commands accepted by `boot`.

---

### Example of problem

```python
mc.boot(boot_delay=10) 
```

If it was intended that this value were used to modify the value in the SV struct this will not happen. Instead the value of `boot_delay` is used to change the delay between transmitting boot packets to the SpiNNaker machine.

After this commit the correct call is

```python
mc.boot(sv_overrides={"boot_delay": 10})
```

### Tests

This is hard to test well, so I've opted for not adding further tests. I can probably be persuaded to change my mind.

Thanks @lplana for spotting this!

### DIFF
--- a/rig/machine_control/boot.py
+++ b/rig/machine_control/boot.py
@@ -62,7 +62,7 @@ spin5_boot_options = {
 def boot(hostname, boot_port=consts.BOOT_PORT,
          scamp_binary=None, sark_struct=None,
          boot_delay=0.05, post_boot_delay=2.0,
-         **sv_overrides):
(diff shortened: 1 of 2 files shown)


neworderofjamie:
Ick, but it makes sense

PR #266, opened 2016-11-23
Indicate if the tag is in use when allocating SDRAM fails

mundya:
Allocating SDRAM can fail either because there is insufficient memory available or because the requested tag is already in use. This commit modifies the `sdram_alloc` method so that it reads the allocation table to determine the cause of the failure and reports this in the error that is raised.

This should make diagnosing memory allocation failures easier.

In addition some Flake8 fixes, and Flake8 threw up a fault in a test which I *might* have fixed correctly. @mossblaser would you mind glancing at the last commit if you have time?

### DIFF
--- a/rig/machine_control/consts.py
+++ b/rig/machine_control/consts.py
@@ -109,6 +109,7 @@ class SCPReturnCodes(enum.IntEnum):
     p2p_timeout = 0x8e  # Eth chip <--> destination comms timeout (Fatal)
     pkt_tx = 0x8f  # Pkt Tx failed (Fatal)
 
+
 RETRYABLE_SCP_RETURN_CODES = set([
     SCPReturnCodes.sum,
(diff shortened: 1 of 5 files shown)


mossblaser:
Sounds like a good idea! Seems like perhaps the wrong fix (I.e. would be better if scamp reports this) but being pragmatic and all that...

Implementation looks sane but obviously I can't verify it actually works!

mundya:
> Sounds like a good idea! Seems like perhaps the wrong fix (I.e. would be better if scamp reports this) but being pragmatic and all that...

Yeah, I was wondering about that. Might talk to ST.

mundya:
Actually, I guess this can be replaced with a SCAMP implemented equivalent later...

mundya:
Rebased onto #267.

